PR #260, opened 2026-03-05
Fix stall detection to allow failure detection on first working rep

9thLevelSoftware:
## Summary
This PR fixes stall detection logic to properly detect failures (e.g., failed bench press) that occur during the first working rep, while still protecting against false positives during brief lockout pauses.

## Key Changes
- **Updated stall detection deferral logic**: Changed `shouldDeferStandardSetStall()` to only defer stall detection when `workingReps == 0 AND !hasPendingRep`. Previously, it deferred whenever `workingReps == 0`, which prevented failure detection when a rep was pending at the top of the first rep.
- **Clarified failure scenario handling**: A stalled pending rep at TOP is a legitimate failure scenario (e.g., failed bench press) and should trigger stall detection, not be deferred.
- **Updated test cases**: Removed calls to `completeFirstWorkingRep()` and adjusted test expectations to reflect that the first working rep should remain incomplete when simulating failure scenarios.

## Implementation Details
- The velocity hysteresis band (STALL_VELOCITY_LOW=2.5, STALL_VELOCITY_HIGH=10.0 mm/s) provides adequate protection against false triggers during brief pauses, making the `hasPendingRep` guard unnecessary for deferral.
- When `workingReps == 0` but `hasPendingRep == true`, the user is mid-first-rep and must be protected from stall detection deferral to allow proper failure detection.
- Test adjustments ensure the failure scenario (pending rep at TOP during first working rep) is properly validated.

https://claude.ai/code/session_017rGcHJxzHsrPfr6sJLLruP

### DIFF
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt
@@ -540,18 +540,18 @@ class ActiveSessionEngine(
 
     /**
      * Standard set stall guard:
-     * - Ignore fail/stall detection before first confirmed working rep.
+     * - Defer stall detection only when no working reps are confirmed AND no rep is pending.
      *
-     * Issue #256: Removed hasPendingRep guard. Previously, stall detection was deferred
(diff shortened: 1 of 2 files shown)


9thLevelSoftware:
@copilot

Issue #340, opened 2026-04-08
[Bug]: single cable exercise volume is doubled


chiradeepv:
### Platform

iOS

### OS Version

Ios26

### App Version

0.60

### Device Model

iPhone 14

### Bug Description

1. Workout with a single cable exercise counts- e.g., bent over row or reverse lunge 10x60 lb
2. App reports volume as 1200 lb instead of 600 lb

### Steps to Reproduce

1. Workout with a single cable exercise counts- e.g., bent over row or reverse lunge 10x60 lb
2. App reports volume as 1200 lb instead of 600 lb

### Expected Behavior

Only single cable work should not be automatically doubled

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have tested with the latest version of the app

9thLevelSoftware:
Fixed across 3 commits on the `mvp` branch: `c5966d49`, `1a9932f1`, `1d95bc16`.

**Root cause:** 11 locations across 8 files unconditionally multiplied weight by 2 (assuming dual-cable), which doubled volume/weight display for single-cable exercises like Reverse Lunge and Bulgarian Split Squat.

**Fixes applied:**

*Commit 1 — Primary hardcoded ×2 paths:*
- `InsightCards.kt`: Changed `effectiveHeaviestKgPerCable() * 2f` → uses `session.cableCount ?: 1`
- `PortalPullAdapter.kt`: Changed hardcoded `cableCount = 2` → `null` (lets heuristic decide)
- `ExerciseImporter.kt`: Changed unknown sidedness default from `DOUBLE` → `EITHER` (defers to runtime telemetry heuristic)
- Expanded `FORCED_SINGLE_CABLE_EXERCISE_NAMES` with "Reverse Lunge" and "Bulgarian Split Squat" (validated against exercise_dump.json)

*Commit 2 — SessionSummary pipeline:*
- Added `cableCount: Int?` field to `SessionSummary` data class
- Added `cableMultiplier` extension property (`if (cableCount == 2) 2 else 1`)
- Updated `selectSessionSummariesSince` SQL query to include `ws.cableCount`
- Replaced 6 hardcoded `* 2` in: `SmartSuggestionsEngine` (3 sites), `ReadinessEngine` (2 sites), `DashboardComponents` (1 site)

*Commit 3 — HealthIntegration + DRY cleanup:*
[...]

PR #451, opened 2026-05-21
fix(workout): detect single-cable usage from telemetry (#340)

Jdub7575:
## Summary

Fixes remaining cases of doubled volume/weight for single-cable workouts when exercise metadata still indicates dual-cable (e.g. Wide Grip Pulldown with both retractors connected but only one cable active).

Addresses the open edge case in #340 after the earlier `mvp` hardcoded `×2` cleanup.

## Problem

- `calculateSetSummaryMetrics()` treated `preferredCableCount = 2` as authoritative, so telemetry never downgraded to single-cable.
- Idle cable load/ROM noise (e.g. ~2.5× load ratio, not 5×) failed the old heuristic.
- `displayMultiplier` from unified-bar metadata (2) was persisted even when only one cable was active, doubling history weight display.
- Some PR/tagging paths stored `displayMultiplier` as physical `cableCount`.

## Solution

- **`CableUsageDetector`**: single-cable detection from load imbalance (3× peak ratio), ROM imbalance (position delta), and concentric velocity dominance on one cable.
- **`ActiveSessionEngine`**: telemetry overrides dual-cable hints; `displayMultiplier = 1` when resolved `cableCount = 1`.
- **PR metadata**: use `preferredCableCount` / session `cableCount` instead of `displayMultiplier` in `GamificationManager`, `DefaultWorkoutSessionManager`, and `SqlDelightWorkoutRepository`.
- **`HistoryTab`**: use `displayLoadMultiplier()` for consistent weight display.

## Test plan

- [x] `CableUsageDetectorTest` — load, ROM, balanced dual-cable cases
- [x] `DWSMWorkoutLifecycleTest` — dual hint + single-cable telemetry → `cableCount=1`, volume not doubled
- [ ] `./gradlew :shared:testAndroidHostTest -Pskip.supabase.check=true`
- [ ] Manual: single-cable pulldown/row with both retractors attached; confirm volume = `weight × reps` (not ×2)

## Out of scope

- No backfill/migration for sessions already saved with doubled `totalVolumeKg`.

---
Contributed from [@Jdub7575](https://github.com/Jdub7575) — happy to adjust based on review feedback.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/SqlDelightWorkoutRepository.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/SqlDelightWorkoutRepository.kt
@@ -860,7 +860,7 @@ class SqlDelightWorkoutRepository(private val db: VitruvianDatabase, private val
             val newVolume = weightKg * reps
             val exercise = exerciseRepository.getExerciseById(exerciseId)
             val exerciseName = exercise?.name ?: ""
-            val cableCount = exercise?.displayMultiplier
+            val cableCount = exercise?.preferredCableCount
 
             val combinedPhase = "COMBINED"
 
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/domain/usecase/CableUsageDetector.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/domain/usecase/CableUsageDetector.kt
@@ -1,0 +1,81 @@
+package com.devil.phoenixproject.domain.usecase
+
+import com.devil.phoenixproject.domain.model.WorkoutMetric
+
+/**
+ * Detects when a set was performed on a single cable despite dual-cable metadata or
+ * both cables being physically connected (Issue #340).
+ */
+object CableUsageDetector {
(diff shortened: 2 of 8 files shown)


9thLevelSoftware:
Going to decline this one, but not because of your code.  I went down this route with the original app and quickly discovered the sensors on this machine are entirely too janky to handle this approach reliably, causing tons of incorrectly-labeled data.  I'm revisiting the current approach soon to make something easier to manage.

Jdub7575:
Thanks for the review and the clear call-out on sensor reliability — that context helps a lot.

I went down the telemetry path because #340 still had reports after the `mvp` hardcoded `×2` cleanup (including single-cable work with both retractors attached but only one cable active). Totally fair that janky load/ROM noise can mis-label sets at scale; I’ve seen that pattern burn teams before.

I’ll leave this PR closed. When you revisit a metadata-first / easier-to-manage approach for cable count and volume, I’m happy to contribute there instead — especially on exercise defaults, explicit single-cable variants, or user-visible cable mode if that’s the direction.

Appreciate you taking the time to review.

Issue #540, opened 2026-06-14
iOS Italian locale: eccentric percentage UI is clipped in Just Lift


9thLevelSoftware:
## Bug Description

On an Italian-locale iPhone running app version 0.9.1, the Just Lift eccentric percentage / eccentric-load UI is not rendered legibly. The reporter says the eccentric percentage is "not shown well". Screenshot OCR around the affected area is garbled/clipped, with visible sections including **Eccentric Load**, **Echo Level**, **Rep Count Timing**, and options **Hard / Harder / Hardest / Epic**.

## Steps to Reproduce

1. On an iPhone using Italian locale/keyboard, open Project Phoenix app version 0.9.1.
2. Go to **Just Lift**.
3. View the **Eccentric Load** / eccentric percentage controls.
4. Observe the text/layout around the eccentric percentage and related controls.

## Expected Behavior

The eccentric percentage / eccentric-load labels and values should render fully and legibly on Italian iPhone layouts.

## Actual Behavior

The reporter says: `On Italian iPhone, the eccentric percentage is not shown well.`

Screenshot OCR indicates labels in the affected section are clipped/garbled; visible OCR includes:

```text
Eccentric Load
Echo Level
Hard / Harder / Hardest / Epic
Rep Count Timing
te entire one poses for s
```

The garbled line appears to correspond to clipped or overlapped explanatory text/percentage display in the Just Lift eccentric settings area.

## Environment

- App: Project Phoenix mobile app
- Version: 0.9.1
- Platform: iPhone / iOS
- Locale/keyboard: Italian (`it-IT`)
- Screen: Just Lift → Eccentric Load / eccentric percentage controls

## Original Discord Report

```text
<@1502654170031849644> User reports multiple issues with UI and value entry on italian iPhone.  app version 0.9.1
```

Additional screenshot text from reporter:

```text
On Italian iPhone, the eccentric percentage is not shown well.
```

## Screenshots / Attachments

- Screenshot_2026-06-13_205919.png: https://cdn.discordapp.com/attachments/1509574296807407717/1515522413947916449/Screenshot_2026-06-13_205919.png?ex=6a2f4fbc&is=6a2dfe3c&hm=365260bdd2e120128a01fa2d84e50f0e9f03f25c4bc303f5bf1b96b6a022e0bf&
[...]

PR #546, opened 2026-06-14
Fix Rest Timer eccentric percent localization

9thLevelSoftware:
## Summary
- route the Rest Timer eccentric-load value through the shared locale-aware percent label helper
- add an integer percent formatter for Rest Timer slider values that are not always `EccentricLoad` enum entries
- extend the eccentric-load regression coverage for Italian NBSP percent rendering on 5% slider increments

## Verification
- `git diff --check`
- `./gradlew :shared:testAndroidHostTest -Pskip.supabase.check=true --no-daemon --console=plain`
- `./gradlew :shared:compileKotlinIosArm64 :shared:compileTestKotlinIosArm64 -Pskip.supabase.check=true --no-daemon --console=plain`

Related to #540.


9thLevelSoftware:
@codex address comments and failing CIs

PR #548, opened 2026-06-14
Fix Routine Overview eccentric percentage localization

9thLevelSoftware:
## Summary
- Localize RoutineOverview Echo/Eccentric headings with existing Compose string resources.
- Route RoutineOverview EchoLevel chips through `echoLevelLabel(level)` instead of enum `displayName`.
- Add an Int-based eccentric percent label helper so slider readouts use the Italian NBSP percent form (`110 %`) while keeping other locales as `110%`.
- Add regression coverage for the formatter and RoutineOverview source call sites.

## Verification
- PASS: `./gradlew :shared:testAndroidHostTest -Pskip.supabase.check=true --no-daemon --console=plain`
- PASS: `./gradlew :shared:compileKotlinIosArm64 :shared:compileTestKotlinIosArm64 -Pskip.supabase.check=true --no-daemon --console=plain`
- PASS: `bash .github/scripts/forbid-tracked-secrets.sh`
- PASS: `bash .github/scripts/validate-ios-schema.sh`
- PASS: `git diff --check`

## Notes
- Refs #540.
- `./gradlew spotlessCheck` currently fails on unrelated pre-existing files from `origin/main` (for example `HealthConnectReenableGuardTest.kt` import ordering and other baseline formatting violations); this PR's touched files were formatted with `spotlessApply`, with unrelated formatter changes restored before commit.

### DIFF
--- a/shared/src/androidHostTest/kotlin/com/devil/phoenixproject/presentation/screen/RoutineOverviewLocalizationGuardTest.kt
+++ b/shared/src/androidHostTest/kotlin/com/devil/phoenixproject/presentation/screen/RoutineOverviewLocalizationGuardTest.kt
@@ -15,7 +15,16 @@ class RoutineOverviewLocalizationGuardTest {
         dir
     }
 
-    private fun read(relativePath: String): String = File(projectRoot, relativePath).readText()
+    private fun read(relativePath: String): String {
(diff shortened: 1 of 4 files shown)


9thLevelSoftware:
@codex Address comments and merge conflicts

PR #550, opened 2026-06-14
fix(issue-549): refresh Home cycle banner via checkAndAutoAdvance

### LINKED ISSUE #549: Bug: Home cycle banner shows stale workout until Cycles navigation refresh
## Summary
The Home screen cycle banner can show the previous day's workout on initial load instead of refreshing to the appropriate/current day. Navigating to Cycles and then back to Home forces the banner to refresh.

## Repository
Mobile app (`project-phoenix-mp`)

## Reported by
Devil via Discord bug report thread: https://discord.com/channels/1505687028107579553/1515733457081204808

## Environment
- App version: not provided
- Platform/device: not provided
- Area: Home screen / Cycles / cycle banner

## Steps to Reproduce
1. Have a training cycle where the expected Home screen banner should advance to today's workout/day.
2. Open/load the app to the Home screen.
3. Observe the cycle banner shown on initial Home screen load.
4. Tap the **Cycles** button.
5. Navigate back to the Home screen.

## Expected Behavior
The Home screen cycle banner refreshes on initial load and displays the appropriate/current day's workout.

## Actual Behavior
The Home screen initially displays a stale cycle banner from the previous day. The reporter's example: today's cycle banner displayed yesterday's workout until the user manually navigated to Cycles and back.

## Workaround
Tap **Cycles**, then return to the Home screen to force the banner to refresh.

## Attachments
None.

9thLevelSoftware:
## Summary

Home screen's cycle banner was reading the active cycle's `CycleProgress` via a pure `getCycleProgress` call, which never reconciles `lastAdvancedAt` against the current calendar day. `TrainingCyclesScreen` is the only other production caller of `checkAndAutoAdvance`, which is exactly why the reporter's tap-Cycles-then-back workaround refreshed the banner.

Route the Home path through a new `loadHomeCycleProgress` helper that calls `checkAndAutoAdvance` first and falls back to `getCycleProgress` if the result is null. The helper uses the returned `CycleProgress?` directly so the banner renders the post-advance `currentDayNumber` on the first composition after a calendar-day boundary.

Fixes #549

## RCA

The RCA is in #549 (comment: 4702176144):

- `HomeScreen.kt:100-102` reads `getCycleProgress` only; never `checkAndAutoAdvance`.
- `TrainingCyclesScreen.kt:200` is the only production `checkAndAutoAdvance` call site in `commonMain`. Its 60s polling loop (lines 226-231) re-runs it for as long as the user is on the Cycles tab.
- `SqlDelightTrainingCycleRepository.checkAndAutoAdvance` (lines 539-574) writes the new `current_day_number` and `last_advanced_at` back to the `cycle_progress` row; subsequent Home reads pick up the new value.
- `TrainingCycleModels.kt:198-268` and `TrainingCycleModelsTest.kt:67-94` are correct; the bug is purely a missing call site on the Home path.

## Changes

- `shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/HomeScreen.kt`
  - New `internal suspend fun loadHomeCycleProgress(repository, cycle)` helper.
  - `LaunchedEffect(activeCycle)` now calls the helper instead of `getCycleProgress` directly.
- `shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/screen/HomeScreenCycleBannerTest.kt` (new)
  - 4 unit tests covering call order, auto-advance honored, fallback to `getCycleProgress`, and no-advance path.
- `shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/FakeTrainingCycleRepository.kt`
  - `callLog` / `resetCallLog` instrumentation on `getCycleProgress` and `checkAndAutoAdvance` to support call-order assertions. Test infrastructure only; no public repo surface change.

## Out of scope

- `UpNextWidget.kt:46-50` has the same pattern but is not currently rendered in the composable tree; leaving for a separate issue if it is reintroduced.
- Lifecycle resume observer for the Home banner (recommended in RCA "A more robust variant"); not required for the headline bug.

## Tests

`gradle :shared:compileAndroidMain :shared:compileKotlinIosArm64 :shared:compileTestKotlinIosArm64 :shared:testAndroidHostTest -Pskip.supabase.check=true`

- `compileAndroidMain`: clean
- `compileKotlinIosArm64`: clean
- `compileTestKotlinIosArm64`: clean
- `testAndroidHostTest`: **2052 tests, 0 failures, 0 errors, 0 skipped**

New tests:
- `loadHomeCycleProgress calls checkAndAutoAdvance before getCycleProgress`
- `loadHomeCycleProgress uses the auto-advanced day from checkAndAutoAdvance`
- `loadHomeCycleProgress falls back to getCycleProgress when checkAndAutoAdvance returns null`
- `loadHomeCycleProgress returns the persisted value when no auto-advance is due`

GPT-5.5/default retains final merge authority; MiniMax/phoenixworker will not merge.

### DIFF
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/HomeScreen.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/HomeScreen.kt
@@ -98,7 +98,13 @@ fun HomeScreen(navController: NavController, viewModel: MainViewModel) {
     var showOneRepMaxComingSoonDialog by remember { mutableStateOf(false) }
 
     LaunchedEffect(activeCycle) {
-        cycleProgress = activeCycle?.let { cycle -> cycleRepository.getCycleProgress(cycle.id) }
+        // Issue #549: route the Home banner through `checkAndAutoAdvance` so the cycle banner
+        // refreshes to the current day on initial Home load. Previously this called
+        // `getCycleProgress` directly, which is a pure read and never reconciles
+        // `lastAdvancedAt` against the current calendar day. `TrainingCyclesScreen` is the only
+        // other production caller of `checkAndAutoAdvance` — that is exactly why the reporter's
+        // tap-Cycles-then-back workaround refreshed the banner.
+        cycleProgress = activeCycle?.let { cycle -> loadHomeCycleProgress(cycleRepository, cycle) }
     }
 
     LaunchedEffect(Unit) {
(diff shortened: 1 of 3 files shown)


9thLevelSoftware:
GPT-5.5 final merge-gate audit:

- PR head verified: `d4c17265cd6dbc7267e5095f6215ba66026d6697`
- GitHub reports the PR is mergeable/clean and linked with `Fixes #549`.
- All 10 status checks are successful, including Android build, iOS test-target compile, unit/lint/security gates, GitGuardian, and Kilo review.
- I reviewed the two Gemini threads about the fallback `getCycleProgress` call. They are non-blocking optimization suggestions: the fallback intentionally preserves the previous Home behavior for edge/orphan progress states while still calling `checkAndAutoAdvance` first for the reported stale-banner bug. No product/security/test blocker remains.

Resolving the non-blocking review threads and merging with a head-SHA guard.

PR #603, opened 2026-06-28
fix(ble): retry reps notification subscription on GATT busy

9thLevelSoftware:
## Problem

A user reported the app stopped counting reps; toggling Bluetooth off/on fixed it. RCA from their connection log:

- **Conn #1 (18:10:42.092):** "Enabling BLE notifications…" → **8 ms later `[ERROR]` Reps notification error — Write failed: `WriteRequestBusy`** → **zero `REP_RECEIVED` events** for the entire session.
- User toggled Bluetooth → `DISCONNECT` 18:11:09.
- **Conn #2 (18:11:13.480):** same setup, **no busy error**, reps flow normally from 18:11:22 (`source=reps-characteristic`).

Machine was healthy throughout (`faults=None`, temps normal) — this was purely the app's BLE setup.

## Root cause

Kable issues the reps **CCCD descriptor write** (enable-notifications) when the `observe()` flow is first collected, and that write goes **outside** `BleOperationQueue` — so the app's own GATT serialization doesn't cover it. During setup, `startObservingNotifications()` fired three `observe()` collections (reps/version/mode) + firmware reads + `pollingEngine.startAll()` within ~8 ms, while a 34-byte command write was still settling. On Android (API 33+, one GATT write in flight at a time) the reps CCCD write got `ERROR_GATT_WRITE_REQUEST_BUSY`. Kable's default handler propagated it through the flow; the old `.catch {}` logged and let the flow **complete** → reps unsubscribed permanently, **no retry**. Reps is the only rep source on Vitruvian (no standard NUS RX), so the session counted nothing. It's a timing race; the Bluetooth toggle just re-rolled it.

## Fix (`KableBleConnectionManager.startObservingNotifications`, `BleConstants`)

Mirrors the existing `BleOperationQueue.write()` busy-retry pattern:

1. **Self-healing reps subscription** — `observeRepsWithRetry()` retries/resubscribes on transient busy with exponential backoff (100→200→400→800 ms, max 5 attempts). Recovers in ~100 ms automatically instead of needing a manual Bluetooth toggle. Permanent failure surfaces a clear "Rep counting unavailable — reconnect required".
2. **Contention removal** — reps subscribes **first and alone**; firmware reads, non-critical version/mode observes, and the polling read-flood are gated behind the reps CCCD write landing (`repsReady`, fail-open after 2 s), so the critical write normally succeeds on the first try.

## Tests

- `:shared:testAndroidHostTest` BLE package: **177 tests, 0 failures** (4 new — transient-error classifier + backoff curve).
- Mobile-only BLE; no sync-DTO / 1RM / portal-parity surface touched.

## Not verified here

⚠️ **Needs one real-device connect/disconnect cycle** — a BLE timing race can't be reproduced off-hardware. Logic is unit-tested and matches the proven command-write retry pattern, but confirm on the trainer before relying on it.

https://claude.ai/code/session_01L3KXNewdGghVDifCA37ZFd

### DIFF
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/data/ble/KableBleConnectionManager.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/data/ble/KableBleConnectionManager.kt
@@ -10,6 +10,7 @@ import com.devil.phoenixproject.util.BleConstants
 import com.devil.phoenixproject.util.HardwareDetection
 import com.devil.phoenixproject.util.rethrowIfCancellation
 import com.juul.kable.Advertisement
+import com.juul.kable.Characteristic
 import com.juul.kable.Peripheral
 import com.juul.kable.Scanner
 import com.juul.kable.State
@@ -19,6 +20,7 @@ import kotlin.time.Clock
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.TimeoutCancellationException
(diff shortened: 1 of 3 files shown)


9thLevelSoftware:
Addressed in 9a5ce8be:

**CI — iOS Test Target Compile (was failing):** `:shared:compileTestKotlinIosArm64` rejected the backtick test name `...transient (default retry)` — Kotlin/Native disallows `()` in declaration names (Android/JVM tolerated it). Renamed to `...transient by default`. That was the only compile error in the job.

**@gemini-code-assist — release `repsReady` on permanent failure:** Done. `observeRepsWithRetry` now calls the gate callback (`onSettled`) on permanent give-up as well as on success, so polling no longer waits out the full `REPS_SUBSCRIBE_READY_TIMEOUT_MS` when the subscription has already exhausted its retries.

**@gemini-code-assist — guard the backoff shift:** Done. `repsBackoffMs` now uses `coerceIn(0, 30)` on the shift amount to rule out a Long shift wrap on a stray large `attempt`.

Local: `:shared:testAndroidHostTest` BLE package green (19 in the connection-manager class, 0 failures). iOS/Native compile can't run on Windows — relying on the re-triggered CI to confirm `compileTestKotlinIosArm64` now passes.

Issue #632, opened 2026-07-07
5/3/1 week auto-advancement for template cycles


9thLevelSoftware:
Follow-up to the #620 cycle-templates rebuild.

## Context
Template cycles now wire into live %-of-1RM resolution, and `TemplateConverter.convert()` accepts a `weekNumber` (1-4) that selects the 5/3/1 percentage scheme (65/75/85 → 70/80/90 → 75/85/95 → deload 40/50/60, folded through the 90% training max into `RoutineExercise.setWeightsPercentOfPR`).

What's missing: nothing advances the week automatically. A 5/3/1 cycle created today stays on week 1 percentages forever unless recreated.

## Proposed design
- Hook: `ActiveSessionEngine.updateCycleProgressIfNeeded()` / `CycleProgress.rotation_count` — one full rotation of the 4 training days = one 5/3/1 week
- On rotation increment for a 5/3/1 cycle, rewrite each main lift's `setWeightsPercentOfPR` for week `(rotationCount % 4) + 1` (a `RegenerateFiveThreeOneRoutinesUseCase`; keep routine IDs stable so CycleDay FKs hold)
- Needs durable template identity on `TrainingCycle` (e.g. `template_id` column) to know a cycle is 5/3/1 — **schema migration + portal `CycleDto` parity check required** (see monorepo CLAUDE.md cross-project rules)
- After week 4 (deload), Wendler prescribes adding 2.5kg upper / 5kg lower to the training max — decide whether to bump stored 1RM or track a separate TM

## Why deferred
Schema + sync-DTO changes were out of scope for the #620 fix; everything else in the converter is already week-aware.

PR #633, opened 2026-07-07
fix(#620): rebuild cycle templates E2E — root-cause fix, live 1RM/PR wiring, customization

### LINKED ISSUE #620: [Bug]: Default Cycle Workout won’t start
### Platform

iOS

### OS Version

iOS 26.5

### App Version

0.9.3

### Device Model

iPhone 15 Pro Max

### Bug Description

Bug: “Start Workout” button unresponsive after activating workout

### Steps to Reproduce

	1.	Click Create Cycle
	2.	Select the 3-Day Full Body workout
	3.	Connect the Vitruvian to the app
	4.	Activate the workout
	5.	Click Start Workout

### Expected Behavior

The workout begins
Actual: Nothing happens — no response when clicking “Start Workout.”

### Logs/Screenshots

_No response_

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have tested with the latest version of the app

9thLevelSoftware:
Closes #620.

## Root cause (why "Start Workout" did nothing)
A prefix-filter paradox: `TemplateConverter` tags routines `cycle_routine_` to hide them from Daily Routines → `RoutineFlowManager` filters that prefix out of the shared routines StateFlow → `ActiveSessionEngine.loadRoutineFromCycle[Async]` looked up routines ONLY in that StateFlow → lookup always missed → `return false` → UI silently ignored it. Template cycles could **never** start.

## Commit 1 — root-cause fix + all silent failures (shippable alone)
- DB fallback (`getRoutineById`) in both cycle lookups
- Snackbars on load/connection failure; cycle cards render exercise lists via DB-merged routines; explicit "routine data missing" state
- **[HOLD: device validation required]** `BleConnectionManager.ensureConnection`: tapping Start while Connecting/Scanning dropped both callbacks (second independent "nothing happens" path); now restarts with the caller's callbacks, teardown sequenced to avoid racing the new scan

## Commit 2 — template normalization + live %-of-1RM/PR resolution
- 5/3/1 was entirely non-functional (`percentageSets = null` → 1 set, no reps, 0kg, no AMRAP). Now carries real week-1 prescriptions folded through the 90% training max (`setWeightsPercentOfPR = [59, 68, 77]`), AMRAP flagged, `weekNumber` param ready for auto-advancement (#632)
- ALL template exercises opt into `ResolveRoutineWeightsUseCase` (`usePercentOfPR` + `ESTIMATED_1RM` basis): weights resolve fresh at every workout start from VBT 1RM → stored 1RM → PRs, and grow with the user
- Never 0kg: 10kg/cable conservative fallback when no data exists
- Silent day-drop fixed: unresolvable days are kept (empty routine + user-visible warning) instead of vanishing
- 1RM input step (skippable, prefilled) for all templates; ID-keyed storage
- Rest days normalized (PPL +1, 5/3/1 +3); templates relocated `data.migration` → `domain.model`

## Commit 3 — customization
- New editable preview between template pick and 1RM step (add/remove/reorder exercises, sets/reps)
- "Edit Exercises" from cycle-day sheet → RoutineEditor (fixed a third instance of the prefix-filter bug in `MainViewModel.getRoutineById`)

## Verification
- 2,331 shared tests green, incl. 13 new (cycle-load fallback regression + production-template converter suite)
- `assembleDebug` green
- **Device validation checklist (before merge, per HOLD convention):**
  - [ ] #620 repro: create 3-Day Full Body → connect → activate → Start Workout → workout begins
  - [ ] 5/3/1 cycle: main lifts show 3 sets (5/5/AMRAP) with non-zero resolved weights
  - [ ] Double-tap Start while connecting → connection completes, no dead-end
  - [ ] Vee_ hardware; Trainer+ if available

Follow-up filed: #632 (5/3/1 week auto-advancement — needs schema + portal DTO parity).

https://claude.ai/code/session_012QHYF2m7nPwancEeQ7CQoz


9thLevelSoftware:
Review feedback addressed in 604e3190:

**Fixed**
- **Codex P1 (confirmed, most important find)**: `ModeConfirmationScreen` auto-fills config weights from 1RM via `ExerciseConfig.fromTemplate`, so the converter's "configured weight pins absolute" check disabled live scaling for anyone who *entered* their 1RMs — the recommended path. `ExerciseConfig` now carries `userEditedWeight`; only an explicit edit in the config modal pins an absolute weight. Regression test added.
- **Codex P2**: rep edits in the template preview recompute `percentOfOneRm` through a shared `defaultPercentOfOneRmForReps()` rule (also seeds newly added exercises). A new test locks all production template prescriptions to that rule.
- **Gemini**: fallback weight now floors at the 0.5kg machine increment (tiny 1RM could round to 0kg); `FakeExerciseRepository` returns null for blank names instead of fuzzy-matching everything.
- **Kilo warning (BLE)**: fair hit on the comment overpromising. Now a single cancellation path: the new connection coroutine `cancelAndJoin()`s the previous job — waiting for it to fully unwind — before radio teardown and the fresh scan. Deterministic by construction, not by hope. `[HOLD: device validation required]` stands.
- **Kilo minor (Plank/Crunch)**: `suggestedMode == null` is now honored as the bodyweight sentinel — light fixed default, never 70% of a resolved 1RM. Test added.
- **Kilo cleanups**: snackbar copy hoisted to consts; duplicate ID-map builds extracted to `CycleTemplate.exerciseIdsByName()`.

**Not fixed, with rationale**
- **Gemini i18n (×2)**: the surrounding code in these exact files is largely hardcoded English ("UP NEXT", "Rest Day - Take it easy!", "Add to Cycle", "Create Training Cycle"). Extracting only the new strings buys nothing; a repo-wide i18n sweep should be its own issue.
- **Kilo `getRoutineById` suspend concern**: verified — the only caller is `RoutineEditorScreen`'s `LaunchedEffect` (suspend context), and CI compiled green on the previous push.
- **Kilo "collapse the preview sheet"**: creation-time editing was an explicit product decision for this rebuild; revisit with usage data.

Shared suite: 2335 tests green.

9thLevelSoftware:
Codex round-2 addressed in e5712968: same-day duplicate additions are now skipped in the preview (plus index-hardened LazyColumn keys in ModeConfirmationScreen as defense-in-depth), and picked bodyweight exercises keep the `suggestedMode = null` sentinel via `Exercise.isBodyweight` instead of being stamped OldSchool. Full shared suite green.

9thLevelSoftware:
Codex round-3 addressed in 04f77745: edit detection now baselines against the exact config the card displays (map entry or `fromTemplate` fallback) instead of the raw map lookup — a first explicit weight edit is always tracked as user-pinned, regardless of seeding. Suite green.

9thLevelSoftware:
Codex round-4 addressed in 4295fe44 — and it surfaced something real: verified against the bundled catalog, Squat and Good Morning ship with `equipment=[]` (so `Exercise.isBodyweight` returns **true** for them) while Crunch ships with HANDLES. Trusting `isBodyweight` in the picker silently disabled live scaling for added major lifts — the worst failure direction. Preview-added exercises are now always stamped cable (OldSchool): if that's wrong for a genuinely-bodyweight addition, the cost is a visible, skippable 1RM prompt rather than silent no-scaling. The underlying catalog data-quality problem affects every `isBodyweight` consumer (incl. the portal sync flag) and is now tracked in #635. Suite green.

9thLevelSoftware:
Codex round-5 addressed in 7cf4bc46 — confirmed real, and pre-existing (the 5/3/1 path had the same defect; routing all templates through the 1RM step made it universal). The prefill map was mutated in place after composition while `OneRepMaxInputScreen` seeds its fields via identity-keyed `remember`, so values never appeared. Lookups now publish a single immutable snapshot behind a brief loading gate, so the form composes with the final prefill on first composition. Suite green.

9thLevelSoftware:
Codex round-6 addressed in b47e7e75: days whose exercises all fail to resolve are now saved unassigned (`routineId = null`) instead of pointing at an empty startable routine — they land in the existing "No routine assigned" state with the "Assign Routine" repair button, and day numbering is still preserved. Regression test updated. Suite green.

Issue #256, opened 2026-02-26
[Bug]: auto-stop no longer working


Trombonerama:
### Platform

Android

### OS Version

Android 16

### App Version

0.5.0

### Device Model

Samsung flip 5

### Bug Description

The five second de-load timer is not working. It never deloaded on a failed bench press

### Steps to Reproduce

Toggle stall detection on, fail a heavy lift, survive

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have tested with the latest version of the app

9thLevelSoftware:
Patch inbound

Trombonerama:
Working!

PR #653, opened 2026-07-14
fix(#652): reset shared stall timer on completed working rep

### LINKED ISSUE #652: Echo VBT state accumulates across progressive sets and stops later sets unexpectedly (related to #649)
## Summary

New recurrence of premature VBT-related set termination during progressively harder sets. The report is related to resolved issue 649, but the sequence is different: failure appears to accumulate across sets as difficulty increases and may indicate stale or cumulative VBT state or numeric overflow.

## Reporter context

- Discord reporter/context: Devil
- Source thread: 1526367469340459039
- Environment/version: not provided in this follow-up

## Relationship to resolved issue 649

Issue 649 (Echo Hardest AMRAP: VBT verbal encouragement cut off, set ends prematurely on 3rd of 3) was resolved by merged PR 650. Its RCA identified an AMRAP auto-stop interaction that could end a set after a VBT verbal cue even when Auto-End on Velocity Loss was disabled.

This report should be investigated as a new regression or related failure rather than assumed to be the same path: the observed behavior progresses across consecutive sets and may involve cumulative VBT values or overflow. Preserve issue 649 as prior-fix context and compare the new code path against PR 650.

## Reported sequence

1. First set: difficulty set to Hard; all 10 reps completed and VBT did not trigger.
2. Second set: difficulty set to Harder; the set stopped at rep 6.
3. Third set: difficulty set to Hardest; VBT failed after approximately rep 3.
4. The user does not have the setting enabled to stop sets when VBT thresholds are reached.

## Expected behavior

- Each set should evaluate VBT using the current set's valid measurements and configured behavior.
- Increasing difficulty across sets should not cause earlier sets to contaminate later VBT calculations or cause an unconfigured VBT threshold to stop a set.
- A VBT measurement failure should fail safely without prematurely ending the workout set.

## Actual behavior

- Later sets stop progressively earlier as difficulty increases.
- The third set reports VBT failure after approximately three reps.
- The user reports no configuration to stop sets when VBT thresholds are reached.

## Investigation hypothesis (not established root cause)

The recurrence and consistency pattern (third set, approximately third rep) suggests cumulative VBT values, stale per-set state, or a numeric overflow or invalid-value path. Confirm or reject this by tracing per-set reset, accumulation, threshold, and failure handling. Do not treat the hypothesis as proof.

## Acceptance direction for RCA

- Identify whether VBT state, counters, samples, thresholds, or failure flags survive across set boundaries.
- Verify numeric ranges and invalid or overflow handling for the progressive Hard to Harder to Hardest sequence.
- Compare the affected path with the 649/650 fix and prove whether the prior guard is bypassed or incomplete.
- Define a targeted regression test that runs multiple consecutive sets with increasing difficulty and verifies no premature stop when VBT auto-end is disabled.

## Original Discord report

Something similar happened today, on my first set it was set to hard, did all 10 without VBT kicking in. set 2 set to harder, it made it to 6 reps and stopped the set, then 3rd set on hardest and it VBT failed after like 3 reps? I do not have it configured to stop sets when hitting VBT thresholds.

9thLevelSoftware:
Fixes #652

## Root cause

`ActiveSessionEngine.handleRepNotification` already treats a completed working rep as authoritative proof of resumed motion (PR #650) and zeroes `deferAutoStopDeadlineMs` at that boundary. The shared `WorkoutCoordinator.stallStartTime` countdown, armed by both low-velocity motion and firmware `DELOAD_OCCURRED`, was never reset at the same boundary. It was cancelled only when a later sample exceeded `STALL_VELOCITY_HIGH` (10 mm/s), so a countdown started at a turnaround, brief pause, or firmware de-load could survive a subsequent valid rep and later call `requestAutoStop()` -> `triggerAutoStop()` -> `handleSetCompletion()`, prematurely ending a set the athlete is still performing.

## Fix

In `ActiveSessionEngine.handleRepNotification()`, inside `if (repCountAfter > repCountBefore)`, call `resetStallTimer()` immediately alongside `deferAutoStopDeadlineMs = 0L`. One shared reset covers both stall arms because both write to the same `coordinator.stallStartTime`.

## TDD evidence

- RED on main@0f00f4b: new `Issue 652 - completed Echo rep cancels an armed stall timer` in `DWSMWorkoutLifecycleTest` failed with `expected:<null> but was:<1783989200633>`.
- GREEN after the one-line fix.
- Focused suite (DWSM + VbtAutoEnd + VerbalEncouragementAutoStopDefer): 81 tests, 0 failures.
- Full `:shared:testAndroidHostTest` suite: 2442 tests, 0 failures, 0 errors, 0 skipped.
- Issue #256 pending-first-rep stall tests (`deload starts stall timer even with pending rep`, `velocity stall auto-stops with pending rep after timer expires`) remain GREEN.

## What did NOT change

- Stall Detection thresholds (2.5 mm/s and 10 mm/s).
- Per-set reset, warm-up gating, or Echo level handling.
- AMRAP position auto-stop (`shouldRunPositionBasedAutoStop`).
- VBT Auto-End (still gated by `autoEndOnVelocityLoss`).
- Any other file outside `ActiveSessionEngine.kt` and `DWSMWorkoutLifecycleTest.kt`.

Issue #256 protection is preserved: a pending first working rep has not completed, so its timer remains armed and the existing 5-second stall guard still fires. Stall Detection remains independent from `autoEndOnVelocityLoss` - no new VBT/stall coupling introduced.

## Files

- `shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt` - production fix (one new `resetStallTimer()` call).
- `shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/DWSMWorkoutLifecycleTest.kt` - new regression test (RED characterization preserved from the RCA artifact).

Audited baseline: `origin/main@0f00f4b28856bfc3dd6536340390ec01b2c3ea03`
RCA: https://github.com/9thLevelSoftware/Project-Phoenix-MP/issues/652#issuecomment-4964220598

### DIFF
--- a/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt
+++ b/shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt
@@ -1100,6 +1100,12 @@ class ActiveSessionEngine(
             // motion; let normal AMRAP / stall auto-stop resume by zeroing the
             // deadline (the source-of-truth field).
             deferAutoStopDeadlineMs = 0L
+            // Issue #652: the same completed-rep boundary is authoritative for
+            // the shared stall countdown (both DELOAD_OCCURRED and low-velocity
+            // arms write to coordinator.stallStartTime). Without this, a stale
+            // countdown started at a turnaround, brief pause, or firmware de-load
+            // can survive a subsequent valid rep and later auto-complete the set.
+            resetStallTimer()
 
             // Capture rep boundary timestamp BEFORE scoring so scoreCurrentRep()
             // and processBiomechanicsForRep() both see the correct metric window.
--- a/shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/DWSMWorkoutLifecycleTest.kt
+++ b/shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/DWSMWorkoutLifecycleTest.kt
@@ -1205,6 +1205,58 @@ class DWSMWorkoutLifecycleTest {
     }
 
     @Test
+    fun `Issue 652 - completed Echo rep cancels an armed stall timer`() = runTest {
+        val harness = DWSMTestHarness(this)
+        harness.fakeBleRepo.simulateConnect("Vee_Test")
+        harness.dwsm.updateWorkoutParameters(
+            WorkoutParameters(
+                programMode = ProgramMode.Echo,
+                echoLevel = EchoLevel.HARDER,
+                reps = 10,
+                warmupReps = 0,
+                stallDetectionEnabled = true,
+                isAMRAP = false,
+                isJustLift = false,
+            ),
+        )
+        harness.dwsm.startWorkout(skipCountdown = true)
+        advanceUntilIdle()
+        completeWarmupReps(harness, warmupTarget = 3, workingTarget = 10)
+        completeFirstWorkingRep(harness, warmupTarget = 3, workingTarget = 10)
+        advanceUntilIdle()
+
+        harness.fakeBleRepo.emitDeloadOccurred()
+        advanceUntilIdle()
+        assertNotNull(harness.dwsm.coordinator.stallStartTime)
+
+        harness.fakeBleRepo.emitRepNotification(
+            RepNotification(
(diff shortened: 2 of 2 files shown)


9thLevelSoftware:
Rolled into #654

PR #786, opened 2026-09-17
fix(ios): stage Compose resources for simulator target

### LINKED ISSUE #785: iOS simulator aborts when Compose generated resources are requested
## Summary

Exact-tag iOS simulator acceptance for #673 is blocked because the app aborts during the first Compose scene commit when generated Compose resources are requested.

## Exact provenance

- Source tag: `testflight/2026091224`
- Peeled commit: `734483a88aadb27f0ff64fba2e994fbb248e5986`
- Xcode project: `iosApp/PhoenixApp/PhoenixApp.xcodeproj`
- Simulator: iPhone 17 Pro / iOS 26.5 (`iosSimulatorArm64`)

## Deterministic reproduction

1. Check out `testflight/2026091224`.
2. Build `:shared:linkDebugFrameworkIosSimulatorArm64` with JDK 21.
3. Process/copy Compose resources and build `PhoenixApp` for an iOS simulator.
4. In a test-only `MainViewController` fixture, render either:
   - the production `RestTimerCard` with a valid unresolved `DropSetOfferUiState`,
   - the production `DropSetOfferCard` directly, or
   - a single `stringResource(Res.string.drop_set_offer_question)` in `PhoenixTheme`.
5. Install and launch.

Observed: build succeeds, app aborts in under two seconds with `SIGABRT`; the triggered stack terminates from Kotlin unhandled exception through `androidx.compose.ui.scene.BaseComposeScene.setContent`.

Control: a fixture using the same `PhoenixTheme` and hardcoded Material 3 text/cards/buttons renders successfully. This isolates the failure to generated-resource loading/packaging rather than basic Compose or the drop-set reducer.

## Suspected seam

Generated resource accessors use:

`composeResources/projectphoenix.shared.generated.resources/...`

The Xcode `Copy Compose Resources` phase uses/output-paths:

`compose-resources/...`

The app bundle contains the hyphenated directory. Adding a second camel-case copy after build did not by itself clear the abort, so this needs source-level diagnosis rather than an evidence-time workaround.

## Required acceptance

- Add a focused iOS simulator regression that requests a generated string and does not abort.
- Fix the resource packaging/lookup contract.
- Prove the production `RestTimerCard` + `DropSetOfferCard` fixture renders on iOS.
- Capture the native production card and independently review the fix.

## Scope boundary

This is an iOS build/runtime acceptance blocker discovered while validating #673. It is not Trainer+ hardware evidence and does not supersede PR #781 (portal sync mapper fix).

9thLevelSoftware:
## Summary

Fixes #785. Xcode previously copied an optional, pre-staged compose-resources directory and silently continued when it was absent. That left simulator builds with generated accessors pointing at composeResources/... but no matching bundle tree.

The Copy Compose Resources phase now selects iosSimulatorArm64 for simulator SDKs and iosArm64 for device SDKs, copies directly from the matching Gradle processed-resource output into compose-resources/composeResources, and fails the build when staging is missing. A focused host contract test prevents the target/path contract from regressing.

## Verification

- ./gradlew --no-daemon --max-workers=2 :shared:linkDebugFrameworkIosSimulatorArm64 -Pskip.supabase.check=true
- plutil -lint iosApp/PhoenixApp/PhoenixApp.xcodeproj/project.pbxproj
- git diff --check

The full host test could not run because this machine has no Android SDK configured. A full app simulator launch is also not claimed here: the Xcode project currently references the device framework path, so independent production-card evidence remains a follow-up acceptance gate.


9thLevelSoftware:
Addressed review finding in 543476cf: IosComposeResourceContractTest now extracts and decodes the PBX shellScript before asserting the target/path contract, so escaped pbxproj serialization is handled without weakening the checks. Added assertions for target-specific source selection, destination copy, and fail-fast missing-resource behavior. Verification: contract probe passed; simulator/device staging simulation passed; missing-resource failure simulation passed; plutil lint and git diff --check passed. Gradle host test remains blocked by missing Android SDK; simulator framework build remains blocked by pre-existing missing CsvImporter.kt in the workspace.

PR #805, opened 2026-09-19
Harden auth: revoked refresh signs out, refresh is serialized, reinstall doesn't inherit tokens

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 9).

### PR 9: Harden auth: revoked refresh signs out, refresh is serialized, reinstall doesn't inherit tokens
- Depends on: none
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/data/sync/PortalApiClient.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/PortalAuthRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/sync/PortalTokenStorage.kt, shared/src/iosMain/kotlin/com/devil/phoenixproject/di/PlatformModule.ios.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/data/sync/PortalTokenRefreshTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/data/sync/PortalTokenStorageTest.kt
- Traces to: FP-7, FP-4, F-045, F-075, F-077, A-030
- Risk: medium
- Description:
  1. Parse GoTrue `error_code`. Treat a 400 from `grant_type=refresh_token`, or `error_code ∈ {refresh_token_not_found, refresh_token_already_used, session_not_found, session_expired}`, as definitive: `clearAuthWithEvent(SessionExpired)`. Do this in all three places (`PortalApiClient.kt:610-611,674-675`, `PortalAuthRepository.kt:202-203`). 5xx, 429 and network errors stay recoverable.
  2. Route `PortalAuthRepository.refreshSession()` through one `PortalApiClient.refreshIfNeeded()` that holds `refreshMutex`.
  3. Add an auth generation counter in `PortalTokenStorage`. `clearAuth` bumps it, and a refresh-originated `saveGoTrueAuth` drops its write if the generation changed.
  4. iOS: on DI init, if the NSUserDefaults marker `phoenix_install_marker` is absent **and** no local database file exists (a fresh install), clear the Keychain service. Then set the marker. An upgrading user (DB present) just gets the marker set and stays signed in.
- Acceptance: new tests: a 400 refresh with `refresh_token_not_found` clears auth, while a 503 preserves it. A `clearAuth` during an in-flight refresh leaves the user signed out. Both fail with this PR's production change reverted. The iOS marker logic lives in a small common function with injected predicates, so it can be tested on host (fresh install → clear, upgrade → keep). The CI `ios-target-tests-compile` job must be green. Full command: 0 failures beyond the baseline. Revert Gradle dirt.

## Implementation summary
# PR 9 summary — Harden auth: revoked refresh signs out, refresh serialized, reinstall doesn't inherit tokens

Branch: brownfield/62026a5c/pr-9-harden-auth-revoked-refresh-signs-out-re (worktree D:/pwt/pr-9)
Commits (this PR's own): eb89777f, fe036f83, 412129b0 (final, review round 1)

## Files changed
- shared/src/commonMain/kotlin/com/devil/phoenixproject/data/sync/PortalApiClient.kt
- shared/src/commonMain/kotlin/com/devil/phoenixproject/data/sync/PortalTokenStorage.kt
- shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/PortalAuthRepository.kt
- shared/src/iosMain/kotlin/com/devil/phoenixproject/di/PlatformModule.ios.kt (UNCOMPILED)
- shared/src/commonTest/kotlin/com/devil/phoenixproject/data/sync/PortalTokenRefreshTest.kt
- shared/src/commonTest/kotlin/com/devil/phoenixproject/data/sync/PortalTokenStorageTest.kt

## What / why
1. Definitive refresh failures. `PortalApiException` gained a trailing `errorCode: String? = null`, filled from GoTrue's `error_code` in `handleGoTrueResponse`. A new `internal fun isDefinitiveRefreshFailure(error)` returns true for a PortalApiException whose status is 400/401/403, or whose errorCode is one of refresh_token_not_found, refresh_token_already_used, session_not_found or session_expired. Network and timeout errors (null status), 429 and 5xx stay recoverable. The duplicated failure branches in `ensureValidToken` and `forceRefresh` were merged into one private `refreshWithStoredTokenLocked()`. A definitive failure now calls `clearAuthWithEvent(SessionExpired)` (it used to be `RefreshFailed(isRecoverable=false)`). A recoverable failure emits `RefreshFailed(isRecoverable=true)` and rethrows, as before.
2. Serialized refresh. Added `open suspend fun PortalApiClient.refreshIfNeeded(): Result<Unit>`, which wraps `ensureValidToken()` under `refreshMutex`. `PortalAuthRepository.refreshSession()` is now just `apiClient.refreshIfNeeded()`. Its only caller is `restoreSession()`, which already checks expiry. That removed the repository's own 401/403 classification and two unused imports.
3. Auth generation. `PortalTokenStorage` has a private `authGeneration` counter, read and written only under `authLock`. `clearAuth` bumps it, and so does every non-refresh `saveGoTrueAuth` (sign-in/sign-up). Bumping on sign-in goes slightly beyond the plan: it also stops an in-flight refresh from overwriting a newly signed-in account. `saveGoTrueAuth(response, expectedGeneration: Long? = null): Boolean` drops the write and returns false when the generation is stale. The refresh path captures the generation before the network call. If its write is dropped, it returns whatever token is stored now: null after a sign-out, or the new account's token after a sign-in, so a valid new sign-in is not failed as unauthenticated (fe036f83). `refreshToken()` is now private (no other callers), so every refresh goes through the mutex and the generation check. Existing callers are unchanged because of the default argument.
4. iOS reinstall. A common `internal fun resetSecureStorageOnFreshInstall(hasInstallMarker, localDatabaseExists, clearSecureStorage, setInstallMarker): Boolean` lives in PortalTokenStorage.kt. With no marker and no DB it clears. With no marker but a DB present it keeps the session. It always sets the marker. On iOS, `clearKeychainOnFreshInstall()` runs as the first statement of the `platformModule` `module {}` body, so it runs at startKoin before any single exists. Koin singles are lazy and MainViewModel creates the DB before PortalTokenStorage, so a check inside the Keychain single would always see a DB and never fire. The marker is NSUserDefaults `phoenix_install_marker`. The DB check looks for phoenix.db, vitruvian.db, the staging and recovery files via `DatabaseFileContext.databasePath`, and `legacyLibraryRootPath()`. The clear is `KeychainSettings(service).clear()`. Any exception is logged and never blocks start-up.

## Tests
- PortalTokenRefreshTest (MockEngine against a real PortalApiClient):
  - `revokedRefreshToken400ClearsAuthAndEmitsSessionExpired`: 400 with `refresh_token_not_found`. Auth is cleared, isAuthenticated is false, and SessionExpired is emitted.
  - `serverError503DuringRefreshPreservesTokens`: the tokens are kept and the failure has statusCode 503.
  - `signOutDuringInFlightRefreshLeavesUserSignedOut`: the mock handler calls clearAuth before returning a 200 session. The user stays signed out.
  - `refreshErrorCodeIsDefinitiveEvenWithoutStatus400` and `successfulRefreshStoresNewSession`.
- PortalTokenStorageTest: a stale-generation save is dropped after clearAuth, a current-generation save is written, and a sign-in during a refresh is not overwritten. Fresh-install helper cases: fresh → clear, upgrade → keep, marker present → no-op.
- Revert check: with the predicate reverted to 401/403 only and the generation check disabled, 5 of the new tests fail: the 400, error-code, in-flight sign-out and both stale-generation tests.
- Command: `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue`
  - shared: 3789 tests, 2 failures (baseline: IosComposeResourceContractTest x2, needs /bin/sh). app: 52 tests, 0 failures.
- `spotlessCheck` fails on 180+ files this PR doesn't touch. I ran `spotlessApply`: none of this PR's six files changed, apart from a pre-existing `createPortalHttpClient` block in PortalApiClient.kt that this PR didn't edit. The formatting changes were then reverted.
- Full run re-done after fe036f83: same totals (shared 3789 / 2 baseline failures, app 52 / 0).

## Review round 1 (412129b0): all 11 items addressed, see reviews/pr-9.md
- The refresh token and the generation are read together under `authLock` (`refreshTokenWithGeneration()`).
- A definitive failure now clears auth only if the generation is unchanged: `clearAuthWithEvent(event, expectedGeneration)` compares and clears under the lock. A stale 400 that arrives after a sign-out or a new sign-in neither wipes the newer session nor emits a spurious SessionExpired.
- A 400 is definitive only when it carries a parsed GoTrue code: `error_code`, or the legacy `error` field such as `invalid_grant`. A bare 400 from a proxy or CDN stays recoverable. Codes in the definitive set, 401 and 403 are still definitive at any status.
- 10 new MockEngine tests cover:
  - `error_code` parsed from the body on a 422;
  - `invalid_grant`;
  - a bare HTML 400;
  - the 503 event, and 429;
  - a missing refresh token;
  - concurrent refreshes sending one request;
  - a sign-in during a refresh (success path);
  - a stale 400 after sign-out + sign-in, and a stale 400 after sign-out.
- With the fixes reverted, 4 of the new tests fail: bare 400, 422 body code, and both stale-400 races.
- The test class KDoc is refreshed and the long KDoc line is re-wrapped.
- Full run after 412129b0: shared 3799 tests, 2 failures (the IosComposeResourceContractTest baseline); app 52 tests, 0 failures.

## Deviations
- The generation is bumped on sign-in as well as on clearAuth (see above).
- `refreshSession()` with a token present but no refresh token now clears auth with SessionExpired. It used to return a failure and leave the dead session in place.
- The in-flight tests call `refreshIfNeeded()` directly and don't build PortalAuthRepository, whose `refreshSession()` is now one line delegating to it. I skipped the optional setMain repository test (R-6 response).

## Uncompiled iOS changes
PlatformModule.ios.kt could not be compiled on Windows. It relies on `NSUserDefaults.boolForKey`/`setBool(_, forKey=)`, `KeychainSettings.clear()` (opted in with ExperimentalSettingsImplementation), `co.touchlab.sqliter.DatabaseFileContext.databasePath` (already used in iosMain), and the internal `DatabaseFileNames` / `legacyLibraryRootPath()`. The CI `ios-target-tests-compile` job must be checked.

## iOS event note
Definitive failures now emit SessionExpired rather than RefreshFailed(false). A grep of iosApp and iosMain found no consumers of RefreshFailed or isRecoverable.

## Follow-ups noticed (not done, out of scope)
- The user email is still logged at Info (`SyncManager.kt` login/signup lines), and iOS sets no Kermit minSeverity (security.md).
- `SyncManager.logout()` / `signOut()` still don't take refreshMutex. The generation counter covers the resurrection case, but a refresh request can still be on the wire after sign-out.

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 5:05 AM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/805).

PR #782, opened 2026-09-16
fix(safety): persist trainer load uncertainty across restart

9thLevelSoftware:
## Summary
- add a trainer-addressed, generation-safe durable machine-safety hazard record
- restore unresolved load uncertainty across cold application graph construction
- provide explicit matching-trainer reconnect and RESET-only release request flow
- make connection-loss warning distinguish transport acknowledgement from physical unload

## Scope
Source-only safety remediation for #769. Existing mid-set reconnect, teardown, lease, polling, and bodyweight gates remain in place. This does not resume workouts or rebuild routine state as safety recovery.

## Tests
- `:shared:testAndroidHostTest --tests com.devil.phoenixproject.presentation.viewmodel.MainViewModelTest --tests com.devil.phoenixproject.presentation.manager.MachineSafetyHazardCoordinatorTest` — passed (38 tests)
- `git diff --check` — passed

## Safety status
Production-graph tests now use a real MainViewModel with DefaultWorkoutSessionManager/ActiveSessionEngine, FakeBleRepository, and a persistent fake hazard store across a fresh graph. No hardware or real BLE/load testing was performed. GATT/transport acknowledgement is not represented as physical unload; issue #769 remains hardware-unverified/CNR.

Fixes #769

PR #806, opened 2026-09-19
Clear the #782 safety arm row when a set ends through a successful RESET

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 1).

### PR 1: Clear the #782 safety arm row when a set ends through a successful RESET
- Depends on: none
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/MachineSafetyCoordinator.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/DWSMTestHarness.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/InMemoryMachineSafetyHazardRepository.kt (new), shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/MachineSafetyHazardCoordinatorTest.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/testutil/FakeMachineSafetyCoordinator.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/e2e/WorkoutFlowE2ETest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/JustLiftCompletionBehaviorTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/MachineSafetyArmingEngineTest.kt (new)
- Traces to: FP-2, FP-8, F-006
- Risk: medium
- Description: Implement KD-1.
  1. In `MachineSafetyCoordinator`, record the armed (trainerAddress, generation, executionId) when `armBeforeMachineCommand` persists.
  2. Add `suspend fun resolveArmedExecution(executionId: Long): Boolean`. Under the mutex it loads the row for that trainer and deletes it via `deleteIfGenerationMatches` only when:
     - the row is `Loaded`;
     - its `executionId` and generation equal the remembered arm;
     - `uiState` is `Hidden`.

     A visible loss or a newer arm is never cleared.
  3. In `ActiveSessionEngine.launchMachineTeardownReset`'s success branch, after `markTeardownReady(lease)` returns true and `connectedTrainerAddress` equals the armed trainer, `await` `resolveArmedExecution(lease.executionId)`. This must happen **before** `takePendingResetStart`, `takeTeardownReadyContinuation` and `tryStartCurrentAcceptedRetryLive`.
  4. Leave the failure branch, `launchRestoredMachineTeardownReset` and pause unchanged.
  5. In the start path (`:8570-8574`), when `canStartMachine()` is false, have the coordinator surface the stored hazard (re-show the recovery UI) instead of a silent `failStart`.
  6. Rewrite the test at `MachineSafetyHazardCoordinatorTest.kt:69-82`. It currently asserts the bug. The new tests are:
     - an armed, then resolved, execution allows the next start;
     - a stale executionId doesn't clear a newer arm;
     - resolve never clears a `recordConnectionLost` row.
  7. **Make the engine test tier exercise the barrier** (R-1). Today `DWSMTestHarness` passes no coordinator, `DefaultWorkoutSessionManager` defaults it to `null` (`DefaultWorkoutSessionManager.kt:259`), and the gate becomes `?.canStartMachine() ?: true` (`ActiveSessionEngine.kt:8571`). That is why no test caught F-006.
     - Move `FakeHazardStore` out of `MachineSafetyHazardCoordinatorTest` into `commonTest/testutil/InMemoryMachineSafetyHazardRepository.kt`.
     - `DWSMTestHarness` wires a real `MachineSafetyCoordinator(persistMachineArming = true)` over that store **by default**, bound to the fake BLE's connected trainer address. The harness exposes it so tests can assert rows.
     - Flip `fakeMachineSafetyCoordinator`'s default (androidHostTest) to `persistMachineArming = true` as well.
     - Tests that deliberately need no barrier opt out explicitly.
- Acceptance: new tests pass with the fix, **and each fails when step 3 is reverted**:
  - (a) In `MachineSafetyArmingEngineTest`, running on the default `DWSMTestHarness`, a routine with two consecutive machine sets reaches `Active` for both. This is the user-mandated two-machine-sets test.
  - (b) In `JustLiftCompletionBehaviorTest`, on the default harness, a Just Lift auto-start successor reaches `Active`.
  - (c) Stop-set, then start again, succeeds.
  - (d) A teardown whose `stopWorkout()` fails or disconnects leaves the row. The next start is then refused **and** the recovery UI becomes Visible.
  - The whole existing engine suite passes with arming ON through the harness. Later PRs 6, 7, 13 and 17 inherit this.
  - Run `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest`. There must be no failures except the 2 known Windows `IosComposeResourceContractTest` failures.
  - Revert `gradle.properties`, and delete `gradle/gradle-daemon-jvm.properties` if the run created it.

## Implementation summary
# PR 1 summary: clear the #782 safety arm row when a set ends through a successful RESET

Branch: `brownfield/62026a5c/pr-1-clear-the-782-safety-arm-row-when-a-set` (worktree D:/pwt/pr-1)
Commits for this PR: `5f8c5f44`, `6395e211` (review round 1)

## What changed and why (KD-1, F-006)
- `shared/src/commonMain/.../presentation/manager/MachineSafetyCoordinator.kt`
  - Remembers the hidden arm it persisted (`armedDocument`). This is set in `recordMachineSessionArmed` only when `showRecoveryUi == false`, and **nulled** when a visible loss is recorded. So a dismissed `recordConnectionLost` row can never be resolved.
  - New `suspend fun resolveArmedExecution(executionId): Boolean`. It runs under the mutex and deletes the row via `deleteIfGenerationMatches` only when all of these hold:
    - a remembered arm exists with the same executionId;
    - `uiState` is `Hidden`;
    - `transport.connectedTrainerAddress` equals the armed trainer. The plan put this check in the engine; it lives in the coordinator instead, where the transport is.
    - the stored row is `Loaded` with the same generation and executionId.
    - Cancellation is rethrown, and other exceptions return false (the barrier is kept).
  - New `suspend fun surfaceStoredHazard()` shows the most recent stored hazard. `restoreOnStartup()` now delegates to it, so the logic is unchanged.
- `shared/src/commonMain/.../presentation/manager/ActiveSessionEngine.kt`. Two lines were added:
  - `launchMachineTeardownReset` success branch: right after `markTeardownReady(lease)` returns true, and before `takePendingResetStart` / `takeTeardownReadyContinuation` / `tryStartCurrentAcceptedRetryLive`, it awaits `machineSafetyCoordinator?.resolveArmedExecution(lease.executionId)`.
  - The start gate: when `canStartMachine()` is false, it calls `surfaceStoredHazard()` before `failStart`, so a refused start is no longer silent.
  - The failure branch, `launchRestoredMachineTeardownReset` and pause are unchanged.
- Test tier (R-1):
  - New `commonTest/testutil/InMemoryMachineSafetyHazardRepository.kt`. It replaces both `FakeHazardStore` (coordinator test) and androidHostTest's duplicate `FakeMachineSafetyStore`, which was deleted; the `MainViewModelTest` FQNs were repointed.
  - `DWSMTestHarness` wires a real `MachineSafetyCoordinator(persistMachineArming = true)` **by default** over that store. It is bound to the fake BLE's connected address through the production `BleRepositoryMachineSafetyTransport(fakeBleRepo)`. It exposes `machineSafetyCoordinator` and `machineSafetyStore`. Two new ctor params were added before the trailing `onPostSaveComputed` lambda:
    - `machineSafetyBarrier: Boolean = true` (opt-out);
    - `machineSafetyStore` (to model a relaunch).
  - `fakeMachineSafetyCoordinator` (androidHostTest) now defaults to `persistMachineArming = true`.

## Tests
- New `MachineSafetyArmingEngineTest` (default harness, arming ON):
  - (a) A routine with two consecutive machine sets reaches `Active` for both. The test asserts exactly one arm row for set 1, the row cleared right after set 1's teardown, and exactly one row for set 2 (a new executionId, set index 1).
  - (c) Stop set, then start again, succeeds, and the row is cleared in between.
  - (d) Two variants: `stopWorkout()` fails, or it disconnects during the write. Either way the teardown goes to RecoveryRequired, the row stays, and the UI stays Hidden. A relaunched harness sharing the store then has its start refused: no command is sent, the state is Idle, and the recovery UI is `Visible` for that execution's row.
- (b) `JustLiftCompletionBehaviorTest.three consecutive handle started sets…` now asserts one arm row per auto-started successor, and that the row is empty after each completion.
- `MachineSafetyHazardCoordinatorTest`: the bug-asserting test was replaced. The tests now cover:
  - armed → resolved → next start allowed;
  - an unresolved row survives a relaunch and the new process can't resolve it;
  - a stale executionId never clears a newer arm;
  - resolve requires the armed trainer to be connected;
  - resolve never clears a `recordConnectionLost` row, including after `hideTemporarily()`;
  - `surfaceStoredHazard` shows the hidden row.

### Revert evidence
- **Step 3 reverted:** only the resolve line removed. 8 failed:
  - (a) and (c) in `MachineSafetyArmingEngineTest`;
  - 6 tests in `JustLiftCompletionBehaviorTest`, including (b).
- **Step 5 reverted:** only the `surfaceStoredHazard` line removed. Both (d) variants failed. (d) cannot fail on a step-3 revert, because step 3 only clears rows and (d) is about a row that must stay. So (d) is tied to step 5 instead. This deviates from the plan's wording "each fails when step 3 is reverted".
- **Suite-wide:** with the harness wired (arming ON) but no production fix, `:shared:testAndroidHostTest` had 3778 tests and 86 failed. By class:

  | Test class | Failures |
  |---|---|
  | DWSMWorkoutLifecycleTest | 39 |
  | Issue687StaleWorkSuppressionTest | 13 |
  | Issue687WorkoutExecutionIsolationTest | 9 |
  | JustLiftCompletionBehaviorTest | 6 |
  | ActiveSessionEngineIntegrationTest | 5 |
  | WorkoutExitPersistenceTest | 5 |
  | WarmupProgressionTest | 2 |
  | WorkoutTeardownRecoveryTest | 2 |
  | IosComposeResourceContractTest (known) | 2 |
  | DropSetCompletionCaptureTest | 1 |
  | Issue673SetEndReasonLifecycleTest | 1 |
  | WorkoutMachineTeardownTest | 1 |

  The engine tier now reproduces F-006.

### Final run
- shared: 3787 tests, 2 failed: `IosComposeResourceContractTest` xcodeResourceStagingFailsWhenProcessedResourcesAreMissing and xcodeStagesResourcesForSimulatorAndDeviceTargets. These are the known Windows /bin/sh failures.
- androidApp: 52 tests, 0 failed.
- `gradle.properties` was reverted, `gradle/gradle-daemon-jvm.properties` was deleted, and the worktree is clean.

## Deviations / test adjustments (all test-only)
- **Opt-outs (`machineSafetyBarrier = false`), each with an inline reason.** In every case a start of B over an armed A with no RESET teardown is correctly refused by the barrier:
  - `Issue687WorkoutExecutionIsolationTest`, 2 cases. Test hooks start B over live A.
  - `WorkoutTeardownRecoveryTest.stale reset cannot discard a replacement teardown continuation`. It installs B while A's reset is in progress.
  - `DWSMWorkoutLifecycleTest.Just Lift reset releases its claim when cleanup throws`. An injected throw happens before any teardown, so the row legitimately stays.
- **Resume authorization added:** `DWSMWorkoutLifecycleTest` `interrupted retry reconnect preserves the absolute occurrence overlay` and `failed execution claim preserves an interrupted workout start override for retry` now call `machineSafetyCoordinator?.authorizeInterruptedWorkoutResume()` before `dwsm.reconnectInterruptedWorkout()`. This mirrors production `MainViewModel.reconnectInterruptedWorkout` (line 745). The engine tier never had it because it had no coordinator.
- **Fixture change:** in `ActiveSessionEngineIntegrationTest.completeLoadedCycleSet`, rest autoplay had already started the next set, and the helper then issued a second `startWorkout` over it (a replacement start). It now skips enter/start when the state is already `Active`. The cycle-completion assertions are unchanged.
- **`WorkoutFlowE2ETest.kt` not edited:** it is covered by the `fakeMachineSafetyCoordinator` default flip and passes with arming ON.
- **Trainer check location:** the "connected to the armed trainer" check is inside `resolveArmedExecution`, not in the engine.
- **No iOS changes.**

## Review round 1 (commit 6395e211). All 20 items are addressed; responses are in reviews/pr-1.md

- **Ordering (R-4/7/16/19).** `resolveArmedExecution` now runs **before** `markTeardownReady`. It is gated on `executionGuard.captureMachineTeardownLease()` still being this lease, and wrapped in `NonCancellable`. The guard stays TearingDown until the row is gone, so no successor, continuation or user start can see the stale row. This deliberately moves the call from the plan's step-3 position (after `markTeardownReady`), as the reviewers suggested.
- **Locking and surfacing (R-2/3/7/11/16).** `canStartMachine` and `surfaceStoredHazard` take the coordinator mutex. `surfaceStoredHazard` is a no-op while the UI is Visible. It takes `liveExecutionId` (the engine passes `outgoingLease`) and does not surface a refusal whose only row is the replaced live set's own hidden arm.
- **Cross-trainer (R-5).** A visible loss only forgets the remembered arm for the same trainer.
- **UI exit for real hazards (R-15, major).** `ConnectionLostDialog` has a new `onAcknowledgeUnloaded` action. It is shown only while a safety hazard is Visible and calls `acknowledgeMachineSafetyUnloaded(generation)`.
  - Copy: "handles at rest, cables not pulling beyond their normal light resting tension". It never claims zero load.
  - It is available at any time while Visible, not only after the RESET was sent, so a powered-off trainer can't lock the user out.
  - Loss recording is unchanged.
  - The dialog's "Follow Vitruvian safety guidance" became "Follow the trainer's safety guidance" (detachment rule).
- **Semantics (R-1/8/17).** Accepted per KD-1: a set resumed after a dismissed loss re-arms over the loss row, and its clean, connected RESET resolves it. The class KDoc names this as the single transport-ACK exception.
  - **Correction to 5f8c5f44's message:** "a recorded connection loss is never resolvable, even after dismissal" holds only for a direct resolve.
- **Tests added:**
  - engine:
    - the gated-delete ordering test (R-10);
    - the replacement-start characterization (R-3/11);
    - an in-process start refused after a failed teardown (R-18);
    - after a relaunch: refused, then Visible, then acknowledge, then Active (R-15);
  - coordinator:
    - the gate waits for an in-flight resolve;
    - surfacing is a no-op while Visible;
    - the live-arm skip;
    - a surfaced row can't be resolved;
    - a resumed loss is resolved;
    - a loss on another trainer keeps this trainer's arm resolvable;
    - an acknowledged loss allows the next start;
  - the cycle fixture asserts the requested exercise and set indices (R-9/12);
  - the Just Lift retry test asserts the row's lifecycle (R-13);
  - the E2E coordinator is bound to `FakeBleRepository` through `BleRepositoryMachineSafetyTransport` (R-6/14/20).
- **Mutation checks (each run on `*MachineSafety*`, each killed by exactly one test):**

  | Mutant | Failing test |
  |---|---|
  | Resolve at the plan position | the gated ordering test |
  | Resolve after continuations | the gated ordering test |
  | Surfacing without `liveExecutionId` | the replacement test |
  | `canStartMachine` without the lock | the gated coordinator test |
  | No Visible early return | the visible no-op test |

- **Final run:** `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue`
  - shared: 3795 tests, 2 failed (the known `IosComposeResourceContractTest` ×2).
  - app: 52 tests, 0 failed.
  - `gradle.properties` was reverted and the daemon JVM file removed.
- **No iOS-only code changed.** The dialog is shared Compose (commonMain) and is compiled for Android here.

- **Replacement start over a live armed set is still refused.** It is now refused without a false warning, and the behavior is pinned by a test. If a product path ever needs a live replacement start, the barrier would block it.
- **Refusals by the teardown guard are not surfaced.** A start refused by `MachineTeardownState.RecoveryRequired` (the in-process failed-teardown case) is rejected before the safety gate. So step 5's recovery UI appears only once the guard is Ready, for example after a relaunch. The existing teardown-recovery UX covers the in-process case.
- **Idle-disconnect hazards (from R-15).** Decide whether an idle unexpected disconnect with no armed row should record a hazard at all. Today every unexpected disconnect records one, which is conservative; the user can now clear it.
- **Row-specific copy (from R-15).** Rows that come from a failed RESET or a relaunch could get their own copy, instead of "Connection Lost".

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 5:05 AM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/806).

PR #808, opened 2026-09-19
Delete the buffered backup importer and keep only the streaming importer

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 21).

> **Stacked PR.** Based on `brownfield/62026a5c/pr-2-enforce-foreign-keys-in-the-test-db-and` (PR 2). Merge that first; GitHub retargets this PR automatically.

### PR 21: Delete the buffered backup importer and keep only the streaming importer
- Depends on: PR 2
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/util/DataBackupManager.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/util/DataBackupManagerRoutineNameTest.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/util/DataBackupManagerRoutineExerciseDropSetTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/util/BackupSerializationTest.kt
- Traces to: FP-3, F-055
- Risk: medium
- Description: Route every import through the streaming importer (it handles small files). Delete the non-streaming import path (`DataBackupManager.kt:600-620` and its helpers) and the size switch (`:197`). Where the two paths differed, keep the streaming behavior, except for any fix documented as present only in the buffered path (e.g. `:1456,1610` comments). Port those first.
- Acceptance: the existing "streaming vs buffered parity" tests become single-path tests and pass. A small-file backup round-trip test passes through the streaming path. Grep: one import entry point. Full command: 0 failures beyond the baseline. Revert Gradle dirt.

## Implementation summary
# PR 21 summary: delete the buffered backup importer and keep only the streaming importer

Branch: `brownfield/62026a5c/pr-21-delete-the-buffered-backup-importer-and` (worktree D:/pwt/pr-21, stacked on PR 2)
Traces to: FP-3, F-055

## Commits (this PR's own)
- `42765ffc` refactor(backup): keep only the streaming importer (F-055)
- `62009230` fix(backup): make the single importer atomic and surface skipped rows (review round 1)

## Review round 1 (reviews/pr-21.md: 4 major, 5 minor, 5 nit; all addressed except R-10, which is wontfix)
- **Atomic restore (R-1/R-4/R-5/R-11):** the entire stream walk, the deferred fix-ups and the profile adoption now run in ONE outer `database.transaction`. The old per-section `committedTransaction` blocks are now plain `run {}` blocks inside it. Any abort rolls the DB back to its pre-import state, as the buffered importer did, so a retry starts clean. Memory stays flat because rows are still streamed; only the SQLite journal grows. The `databaseWorkCommitted`/normalize-in-catch logic became dead code and was removed. As a consequence, a failure no longer calls reconcile, because nothing was committed. `BackupJsonNavigatorTest` `...post-identity failure rolls back everything` now asserts the profiles are unchanged and 0 reconcile calls.
- **Counted skips instead of aborts (R-11):** `insertRoutine`, `insertSupersetIgnore`, `insertTrainingCycle`, `insertCycleProgressIgnore`, `insertCycleProgressionIgnore`, `insertPlannedSetIgnore`, `insertCompletedSetIgnore`, `insertProgressionEventIgnore` and `insertStreakHistoryIgnore` are now wrapped in `tryImport`. Counts and id sets update only on success. `insertUserProfileIgnore` stays unwrapped (identity invariant).
- **Deferred superset refs (R-7/R-12):** queued only when the insert actually created the row (`QueryResult.value > 0`).
- **Validation (R-2/R-3/R-6/R-13):** a file with no `data` object is rejected with the friendly malformed message. So is a file where a section had rows but none decoded (new `decodeRow` helper tracks rows seen and decoded per section). The restore dialog (SettingsTab) shows "Records that could not be restored (skipped): N" when `entitiesWithErrors > 0`. The new string `import_records_unreadable` is only in `values/strings.xml`; other locales fall back to English.
- **New tests (DataBackupManagerRoutineNameTest):**
  - `mid-stream failure leaves the database unchanged and a retry restores everything` (truncated file, then good file: metric, completed set, group and superset links, routine name)
  - `section whose rows all fail to decode fails the restore and changes nothing`
  - `json without a data object is rejected as malformed`
  - `non-integer version is reported with the friendly malformed message` (R-9)
  - `dangling group and superset links stay null and an FK miss is a counted skip` (R-8/R-11)
  - the per-row test now uses one good and one bad row.

  With the outer transaction replaced by `run {}`, 2 tests fail (the mid-stream test and the all-undecodable test).
- **R-10:** wontfix. `AndroidDataBackupManager` needs a real Context; no Robolectric harness is added here.
- **R-14:** carryover handled by the orchestrator.
- **Full suite after round 1:** `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue` gave shared 3,792 tests (2 skipped, 0 failures) and app 52 tests (0 failures). No Gradle dirt.
- **Diff noise:** re-indenting the importer body under the outer transaction inflates the diff. `git diff -w` for `DataBackupManager.kt` in `62009230` is +120/-109.

## Files changed
- `shared/src/commonMain/kotlin/com/devil/phoenixproject/util/DataBackupManager.kt`: deleted `importFromJson` from the interface and its ~720-line implementation, plus `STREAMING_IMPORT_THRESHOLD`. Ported the buffered-only fixes into `importFromStream` (see below). Updated the KDoc and stale comments.
- `shared/src/commonMain/kotlin/com/devil/phoenixproject/util/BackupJsonNavigator.kt`: `parseError` now throws `MalformedBackupException`, a new subclass of `IllegalStateException`, so existing callers and tests keep working.
- `shared/src/androidMain/.../DataBackupManager.android.kt`: `importFromFile` always streams. Deleted the now-unused `getFileSizeOrNull`.
- `shared/src/iosMain/.../DataBackupManager.ios.kt`: `importFromFile` always streams. Removed the unused `dataWithContentsOfFile` import. **Not compiled (iOS can't be built on this host).** The change is a straight deletion of the size branch. The other Foundation imports are still used elsewhere in the file.
- `shared/src/commonTest/.../testutil/FakeDataBackupManager.kt`: removed the `importFromJson` override. `importFromFile` returns the empty result directly.
- `shared/src/androidHostTest/.../DataBackupManagerRoutineNameTest.kt`, `BackupJsonNavigatorTest.kt`: tests changed and added (see below).

Not changed, although the plan lists them: `DataBackupManagerRoutineExerciseDropSetTest.kt` and `BackupSerializationTest.kt`. Neither calls an importer; they only test (de)serialization, so nothing needed to change.

## Divergence audit (buffered vs streaming) and what was done
Fixes present only in the buffered importer, now ported to streaming:
1. **Routine-name resolution** (`resolveImportedRoutineName`: "Just Lift" for Just Lift sessions, lookup by `routineId`, inference from a unique exercise). Streaming only sanitized `session.routineName`, and its KDoc called this "simplified". Sessions come before routines in the file, so the fix is a post-stream pass. The importer keeps the parsed routines and routine exercises (small) and a compact `SessionRoutineNameRef` per inserted session. It then calls `updateSessionRoutineName` wherever the resolved name differs.
2. **Routine groups before routines.** `Routine.groupId REFERENCES RoutineGroup` is a foreign key, and exports write `routineGroups` last. The streaming comment claiming "Routine.groupId has no FK constraint" was wrong. With FKs on, which is the production setting, any ≥50 MB restore containing a grouped routine threw from `insertRoutine` and aborted the whole import. The importer now inserts `groupId` only when the group is already known. Otherwise it defers the link and sets it with `updateRoutineGroupId` after the stream. If the group never arrives, the link stays null.
3. **Supersets before routine exercises.** `RoutineExercise.supersetId REFERENCES Superset`, and exports write `supersets` after `routineExercises`. In streaming, such rows hit the FK, were caught by `tryImport`, and were **silently dropped**. The same deferral now applies, fixed with `updateExerciseSupersetRef(supersetId, orderInSuperset, id)`.
4. **`routineGroupsImported` / `routineGroupsSkipped`** were missing from the streaming `ImportResult` and are now set.
5. **Friendly malformed-file error.** Buffered wrapped decode failures in `IllegalArgumentException("Backup file is malformed or produced by an incompatible app version (...)")`. Streaming now does the same for structural parse errors (`MalformedBackupException`, `NumberFormatException`).

Fixes present only in streaming, kept as they are: the `:1456` routineGroups case (F068) and the `:1610` F304 cable count. Both were already in streaming, and the buffered path had the equivalent.

Other differences where the streaming behaviour is kept, as the plan says:
- (Superseded in round 1: the restore is now one all-or-nothing transaction again.)
- A row that fails to decode is skipped and counted in `entitiesWithErrors`; it no longer fails the whole file. The old `malformed top-level JSON` test used `[{}]`, which is now a per-row skip. That test now uses a truncated file, and a new test pins the per-row skip.
- `importedRoutineExerciseIds` only includes rows actually inserted, so it's stricter.
- `sessionsAdopted` log count differs slightly (logging only).

## Tests
- Parity tests became single-path tests: `buffered and streaming preference restores…` became `preference restore writes every typed section for every profile`. The completed-set import tests (end reason, negative/zero attempt) were merged. All 26 former `importFromJson` call sites now go through `importFromStringStreaming` (streaming), and the tests were renamed without the "buffered"/"importFromJson" prefixes.
- New `small backup round trip restores forward references through the streaming importer` (DataBackupManagerRoutineNameTest): a backup whose section order matches the export (forward references) is imported, re-exported with the real streaming exporter, and restored into a fresh DB. The test asserts `groupId`, both `supersetId`/`orderInSuperset`, `routineGroupsImported == 1`, 0 entity errors, and the "Push Day" / "Just Lift" routine names.
- New `well-formed file with an undecodable row skips that row and imports the rest`.
- Revert check: with the post-stream fix-up disabled, 3 tests fail (both routine-name tests and the new round trip). Before this PR, the streaming path failed the round trip on the routine-group FK.
- Targeted: `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest --tests '*DataBackupManagerRoutineNameTest*' --tests '*StreamingImportRoundTripTest*' --tests '*BackupJsonNavigatorTest*'` gave 42 + 12 + 11 tests, 0 failures.
- Full: `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue` gave shared 3,787 tests (2 skipped, 0 failures) and app 52 tests (0 failures). The 2 baseline `IosComposeResourceContractTest` failures didn't appear in this run (reported as skipped/not failing). No regressions.
- Grep: `importFromJson` and `STREAMING_IMPORT_THRESHOLD` have 0 hits. `importFromStream(` has one definition, called only from the Android and iOS `importFromFile`.
- Gradle dirt: `gradle.properties` was not modified, and no `gradle-daemon-jvm.properties` was created.

## Follow-ups noticed (not fixed, out of scope)
- (Round 1: the unwrapped FK inserts are now counted skips, and the restore is atomic.) PR 22 still needs to restore Exercise rows before rows that reference them, so those rows are not skipped on a fresh device.
- The metric "batching" loop in the importer no longer has a transaction effect (it only drives progress callbacks). It could be simplified later.
- Streaming still depends on sections arriving in dependency order: routineExercises after routines, plannedSets after routineExercises, completedSets after sessions, and cycle children after cycles. Every exporter writes that order; hand-edited files in another order lose child rows (logged).
- `exportAllData`/`exportToJson` (the buffered *export*) are still there; `getShareableContent` uses them. That's outside this PR.
- `updateSessionRoutineName` does not bump `updatedAt` (F-057). This is harmless for fresh restores, whose `updatedAt` is null.

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 5:04 AM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/808).

Issue #761, opened 2026-09-08
[TestFlight] Just lift only allows one set to be completed. Upon trying t — iPad14_1


9thLevelSoftware:
## TestFlight Screenshot Feedback

**Reported via:** TestFlight Screenshot Feedback (App Store Connect API)
**Submission ID:** AMGefW4wxXKKc0TWqhmpg1c
**Date:** 2026-09-08T11:37:46.761Z
**Device:** iPad14_1
**OS:** 26.6.1
**Locale:** en-US
**App Bundle ID:** com.devil.phoenixproject.projectphoenix

### Tester Comment

Just lift only allows one set to be completed. Upon trying to start the second set, it does not sense the handles being lifted. 

### Screenshots

_No screenshots attached_
### Environment

- Architecture: arm64e
- Battery: 100%
- Screen: 744x1133 pts

---
*Auto-filed by Phoenix TestFlight Monitor on 2026-09-08T09:26:26.464556*

9thLevelSoftware:
Duplicate of #756 (Just Lift autostart bug — already fixed in PR #759, merged)

9thLevelSoftware:
Consolidated into #756 — same Just Lift autostart bug. Report added as comment.

9thLevelSoftware:
## Reopened — post-fix recurrence continues on testflight/2026090521

**Submission ID:** APimsFeyh9qY2mOY8vk7kBI
**Date:** 2026-09-09
**Device:** iPhone14_2 | **OS:** 26.5

**Tester comment:**
> After completing a just lift set, you can not start another set. Must disconnect and reconnect trainer to have moving the handle recognized.

This issue was closed but the bug persists on the Sept 5 build (testflight/2026090521) which includes PR #759. Reopening. Cross-references: master issue #756, open duplicate #765.

9thLevelSoftware:
## Post-Fix Recurrence — TestFlight Screenshot (2026-09-09)

**Submission ID:** AJpvIVGafGU9jmgJ-FW7CnU  
**Device:** iPad13_1  
**OS:** 26.6.1  
**Build:** testflight/2026090521 (includes PR #759 Just Lift autostart fix)  
**Date:** 2026-09-09T13:22:29.713Z

**Tester comment:**
> "Just lift seems to be broken. It will not do the second set of the first that happens then it has a force rest. And the handle say that it's gonna be ready again and you pick up the handles and nothing happens."

[...]

PR #813, opened 2026-09-19
Make Just Lift auto-start level-triggered so a late grab still starts the next set (#761)

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 6).

> **Stacked PR.** Based on `brownfield/62026a5c/pr-1-clear-the-782-safety-arm-row-when-a-set` (PR 1). Merge that first; GitHub retargets this PR automatically.

### PR 6: Make Just Lift auto-start level-triggered so a late grab still starts the next set (#761)
- Depends on: PR 1
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/JustLiftCompletionBehaviorTest.kt
- Traces to: FP-8, F-043, F-048, F-011
- Risk: medium
- Description: Only in the Just Lift *waiting-for-successor* context, i.e. after a completion that called `enableJustLiftWaiting` (`ActiveSessionEngine.kt:11457-11458`): when the summary reset brings the state to Idle, or teardown enters `Ready`, re-read `bleRepository.handleState.value`. If it is `Grabbed` and no countdown is running, call `startAutoStartTimer` with the then-current lease (after the summary reset `currentLease == null`, and `hasAutoStartAuthority(null)` passes, per code.md §Verification). Idle entries caused by MANUAL_STOP, END_WORKOUT, STOP_SET or SKIP_EXERCISE teardowns must **not** arm the level check, so a user who stops while holding the handles is never auto-restarted. Keep the existing edge trigger. Don't change lease semantics.
- Acceptance: new test from the tests.md probe: `DWSMTestHarness` with `autoStartCountdownSeconds = 2` and `summaryCountdownSeconds = 5`. Complete a set, emit `WaitingForRest`, advance 4,000 ms, emit `Grabbed`, advance 15,000 ms. The state must be `Active` with a new lease. It fails with this PR's production change reverted (ends `Idle`). Negative test: stop the Just Lift set manually while the handles stay `Grabbed`, advance 15,000 ms, and the state stays `Idle` with no machine command sent. The existing immediate-grab test still passes. Arming is ON (PR 1 default). Full command: 0 failures beyond the baseline. Revert Gradle dirt.

## Implementation summary
# PR 6 summary: make Just Lift auto-start level-triggered so a late grab still starts the next set (#761)

Branch: `brownfield/62026a5c/pr-6-make-just-lift-auto-start-level-triggere` (worktree D:/pwt/pr-6). It is stacked on PR 1 (`5f8c5f44`, `6395e211`).
Commits for this PR: `30b276a0`, `229fe6f2` (review round 1)

## Files changed
- `shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt` (+23 lines)
- `shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/JustLiftCompletionBehaviorTest.kt` (+3 tests)

### The cause
1. A `Grabbed` edge during a timed summary starts a countdown that is bound to the completed lease.
2. The summary reset then retires that lease, and the countdown aborts at its next tick.
3. The handles stay `Grabbed`, so no new edge ever arrives, and the next set never starts.

### The fix
- **New helper `restartJustLiftAutoStartIfHandlesHeld()`.** It runs only when the params are Just Lift with `useAutoStart` on and `bleRepository.handleState.value == Grabbed`. It then does two things:
  - It cancels any running countdown. At that point the countdown can only be bound to the lease that was just retired, so it could never complete anyway.
  - It calls `startAutoStartTimer(expectedLease = null)`, which runs only while no lease exists. `startAutoStartTimer` still checks that teardown is `Ready`, that the state is Idle or SetSummary, and at the end of the countdown that the handle is still Grabbed and `useAutoStart` is on.
- **Where it is called.** The helper runs only in the Just Lift waiting-for-successor resets, after the state is set to Idle and the egg timer has started. There are three call sites:
  - the completion job's skip-summary reset lambda;
  - the completion job's timed-summary reset lambda;
  - `dismissCompletedJustLiftSummary`, for manual dismissal.
- **Gating.** In the two completion-job sites the call is gated on `completion.reason != USER_STOPPED`. The dismissal site uses the same gate.
- **The trap noted in the plan.** At reset time the stale countdown still occupies `coordinator.autoStartJob`, so "no countdown is running" is false. A plain `startAutoStartTimer` would return early and the test would still end Idle. That is why the helper cancels first.
- **What cannot trigger it.** MANUAL_STOP, END_WORKOUT, STOP_SET and SKIP_EXERCISE teardowns, and public `resetForNewWorkout()`, never reach these lambdas. A user-stopped Just Lift summary that is later dismissed is excluded by the reason gate.
- **Unchanged.** The edge trigger and the lease semantics.

## Tests
New tests in `JustLiftCompletionBehaviorTest` (`DWSMTestHarness`, arming ON by default):
- **`late grab held through timed summary reset still starts the successor`.** Setup: summary 5 s, auto-start 2 s. The test completes a set, emits `WaitingForRest`, advances 4,000 ms, emits `Grabbed` and advances 15,000 ms. It then asserts:
  - the state is `Active` with a new Just Lift lease;
  - the successor holds the single #782 arm row;
  - `stopWorkoutCallCount == 1`.

  **With the production change reverted** (the engine file stashed), this test fails with "expected Active, actual Idle". Every other test passes.
- **`manual stop while holding the handles never auto restarts`.** The test calls `stopWorkout()` with the handles Grabbed, which leads to SetSummary. It then calls `proceedFromSummary()`, which leads to Idle while still Grabbed, and advances 15,000 ms. It asserts:
  - the state stays Idle with no lease;
  - there is no countdown;
  - no new commands were received and no extra stop calls were made.

  **Mutation check:** I removed the `USER_STOPPED` gate at the dismissal call site. The test then fails ("expected Idle"), which proves it guards the "never restart a user who stopped" rule.
- **`ending Just Lift while holding the handles never auto restarts`.** Calls `stopWorkout(exitingWorkout = true)`, advances 15 s and asserts Idle, no lease, no countdown and no new commands.
- **The completion-job gate is untested.** It is defensive only. Today no Just Lift completion carries USER_STOPPED through `handleSetCompletion`: the only such call, `stopAndReturnToSetReady`, is gated on `isRealRoutine`, and `stopWorkout()` goes through `completeStop`. So removing the `awaitingSuccessorGrab` gates at the two completion-job sites would fail no test. Only the dismissal gate is discriminated, by the mutation check above. A reviewer may prefer to drop the completion-job gate.
- **Existing tests.** The existing immediate-grab test (`handle grab during timed summary protects successor…`) and all other Just Lift tests still pass.

### Full run

| Suite | Tests | Failed |
|---|---|---|
| shared | 3798 | 2 |
| androidApp | 52 | 0 |

The 2 shared failures are the known Windows `/bin/sh` baseline: `IosComposeResourceContractTest` `xcodeResourceStagingFailsWhenProcessedResourcesAreMissing` and `xcodeStagesResourcesForSimulatorAndDeviceTargets`.

`gradle.properties` was reset with `git checkout` in case it was dirty, the daemon JVM file was absent or removed, and the worktree is clean.

## Deviations from the plan
- **"Or teardown enters Ready" has no separate hook.** In Just Lift every Ready, including `retryMachineTeardown`, flows into the lambdas above: `teardownReady.await()` comes before the summary is published, the re-arm requires Ready, and the summary reset uses `skipMachineTeardown = true`, so Ready persists. A hook at `markTeardownReady` would also fire for stop/skip teardowns, which the plan forbids.
- **Countdown is restarted unbound.** The plan said "with the then-current lease". The helper passes `expectedLease = null` explicitly, which is strictly safer: if a lease somehow exists, nothing starts.
- **Manual summary dismissal is also covered.** This is the same bug class: a grab during the summary, then the user taps continue, retires the lease under the countdown. It is gated on a non-user-stopped completion.

## Uncompiled iOS changes
None. All changes are in commonMain and commonTest.

- **`Moving` does not trigger the level check.** A user whose detector reports `Moving` (position extended, no velocity) at reset time gets no restart until a `Grabbed` edge. This matches the existing edge path, which also ignores `Moving`.
- **The restarted countdown starts over.** A countdown restarted at the reset begins again from the full `autoStartCountdownSeconds`, so the user sees it jump back. This is cosmetic.

## Review round 1 (commit 229fe6f2). All 10 items resolved; responses are in reviews/pr-6.md
- **Connection check (R-4/R-9).** `restartJustLiftAutoStartIfHandlesHeld()` now returns unless `bleRepository.connectionState.value is ConnectionState.Connected`. The detector keeps `Grabbed` across an unexpected link drop, so without this the helper could re-arm from a stale reading and end in an unrequested NOT_CONNECTED rejection.
- **Unreachable code (R-2/R-6).** The skip-summary call site is removed, along with the `awaitingSuccessorGrab` val. It was a no-op because `enableJustLiftWaitingMode()` has just set `WaitingForRest`. The timed-summary USER_STOPPED gate stays inline, commented as defensive. The helper now has two call sites: the timed-summary reset and manual dismissal.
- **New tests (5).** All are in `JustLiftCompletionBehaviorTest`, which now has 16 tests. Arming is ON throughout.
  - `grab then manual summary dismissal while holding still starts the successor` (R-1/R-3/R-5). It ends Active with a new lease that holds one arm row.
  - `releasing during the restarted countdown never starts a set` (R-8).
  - `auto start off never restarts a countdown for held handles` (R-8).
  - `held handles read before a link drop never restart the countdown` (R-9).
  - `restarted countdown is refused by a stored safety hazard and surfaces it` (R-10). It ends Idle with no commands, and the foreign row is `Visible`.
- **Labelled test (R-7).** The end-workout test is now commented as a regression guard only.
- **Mutation checks.** Each mutant was run on `*JustLiftCompletionBehaviorTest*`.

  | Mutant | Failing tests |
  |---|---|
  | Dismissal call removed | exactly the dismissal-positive test |
  | Connection check removed | exactly the link-drop test |
  | `useAutoStart` check removed | exactly the auto-start-off test |
  | Timed-site call removed | the late-grab, release-during-countdown and stored-hazard tests |

- **Final run:** `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue`

  | Suite | Tests | Failed |
  |---|---|---|
  | shared | 3803 | 2 |
  | androidApp | 52 | 0 |

  The 2 shared failures are the known `IosComposeResourceContractTest` pair on Windows. `gradle.properties` was reset and the worktree is clean.
- **Not done.** No connection re-check at countdown completion inside `startAutoStartTimer`. That would harden the pre-existing edge path too, and the reviewer marked it out of scope. It is left as a follow-up.

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 4:17 AM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/813).

PR #815, opened 2026-09-20
Attempt RESET before disconnect when the workout engine is torn down mid-set

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 7).

> **Stacked PR.** Based on `brownfield/62026a5c/pr-6-make-just-lift-auto-start-level-triggere` (PR 6). Merge that first; GitHub retargets this PR automatically.

### PR 7: Attempt RESET before disconnect when the workout engine is torn down mid-set
- Depends on: PR 6
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/viewmodel/MainViewModel.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/DefaultWorkoutSessionManager.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/presentation/viewmodel/MainViewModelTest.kt
- Traces to: FP-2, F-013, F-051, A-002
- Risk: medium
- Description: In `MainViewModel.onCleared`, inside the existing `NonCancellable` block and before `bleRepository.disconnect()`: if the engine holds a machine lease (the state is not Idle, or teardown is not `Ready`), call `bleRepository.stopWorkout()` with `withTimeoutOrNull(BleConstants.GATT_OPERATION_TIMEOUT_MS)`. Only then call `cleanup()` and disconnect. Never delete the hazard row on this path, because the relaunch warning must stay (A-002 answer: assume load may remain). The application-scoped engine is out of scope (KD-8).
- Acceptance: new test: an active Just Lift or routine set with arming ON, then `onCleared()`. `FakeBleRepository` records `stopWorkout` before `disconnect`, and the hazard row is still present. A second test: a stop timeout still disconnects. Both fail with this PR's production change reverted. Full command: 0 failures beyond the baseline. Revert Gradle dirt.

## Implementation summary
# PR 7 summary: attempt RESET before disconnect when the workout engine is torn down mid-set

Branch: `brownfield/62026a5c/pr-7-attempt-reset-before-disconnect-when-the` (worktree D:/pwt/pr-7). It is stacked on PR 6 (`30b276a0`, `229fe6f2`), which is stacked on PR 1 (`5f8c5f44`, `6395e211`).
Commits for this PR: `0e271dc1`, `6567d9f3` (review round 1)

## Files changed
- `shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/viewmodel/MainViewModel.kt`: `onCleared` changed, plus 2 imports (+35/-4 over both commits).
- `shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/FakeBleRepository.kt`: new `Event.Disconnected`, recorded in `disconnect()` (+2).
- `shared/src/androidHostTest/kotlin/com/devil/phoenixproject/presentation/viewmodel/MainViewModelTest.kt`: 5 new tests and 3 small helpers (+140/-2 over both commits).

## What changed and why (F-013, F-051, FP-2, A-002)
- **The lease check.** `onCleared` first cancels the connection job. It then reads `holdsMachineLease`, which is true when `(workoutState !is Idle || machineTeardownState !is Ready)` **and** the link is still `Connected` (the connected clause was added in review round 1, R-10). The read happens synchronously, **before** `cleanup()`, because `cleanup()` invalidates the lease.
- **The coroutine.** Inside the existing `viewModelScope.launch(NonCancellable)`, the order is:
  1. If a lease is held: `withTimeoutOrNull(BleConstants.GATT_OPERATION_TIMEOUT_MS) { bleRepository.stopWorkout() }`. A confirmed RESET logs at INFO; a failed `Result` or a timeout logs at warn as "NOT confirmed"; exceptions are caught and logged with their stack.
  2. `workoutSessionManager.cleanup()`.
  3. `bleRepository.disconnect()`, which is the same code as before.
- **Why `cleanup()` moved.** It now runs after the RESET attempt, as the plan says. `cleanup()` also stops the workout foreground service, and that service is what keeps the process alive while the RESET write lands. When no lease is held, the launch runs on `Main.immediate` with no suspension before `cleanup()`, so the idle path still runs synchronously as before.
- **Why the ViewModel has to do this.** The engine runs on `viewModelScope` (MainViewModel.kt:600/657). In production, `ViewModelStore.clear()` cancels that scope before `onCleared` is called, so the engine's own teardown can never run at this point.
- **Hazard row is never cleared.** The call is a **raw** `bleRepository.stopWorkout()`. It does not go through the engine's stop or teardown path, which calls PR 1's `resolveArmedExecution` and would delete the #782 arm row. Nothing on this path touches `MachineSafetyCoordinator`, so the row survives and the relaunch warning still fires (A-002: assume the trainer keeps resisting).
- **The only loss recorder can't fire here.** The only `recordConnectionLost` caller (MainViewModel.kt:~1260) collects `reconnectionRequested` on `viewModelScope`. That scope is already dead at `onCleared`, and an intentional `disconnect()` does not emit a reconnection request anyway. So the same arm row survives unchanged.
- **Small reorder.** `bleConnectionManager.cancelConnectionJob()` now runs before the lease read and `cleanup()`; before this change it ran after `cleanup()`. This has no functional effect.
- **Engine and manager files unchanged.** The plan listed them, but the ViewModel already exposes both states it needs, so no new accessor was added.

## Tests (arming ON: MainViewModelTest's coordinator uses the default `persistMachineArming = true`)
The two tests below are the ones the plan's acceptance names; review round 1 added three more (see that section). All of them clear the ViewModel through `ViewModelStore().put(...).clear()`. That is the production order: the scope is cancelled first and `onCleared` runs after, so the tests prove the `NonCancellable` work runs on a dead scope.

1. **`clearing the ViewModel mid Just Lift set sends RESET before disconnect and keeps the arm row`**
   - Setup: an Active Just Lift set holding exactly one arm row.
   - Asserts the event order `StopWorkoutEntered` < `StopWorkoutCompleted` < `Disconnected`.
   - Asserts `stopWorkoutCallCount == 1` and `disconnectCallCount == 1`.
   - Asserts the store still holds exactly the same arm document.
2. **`clearing the ViewModel mid manual set still disconnects when RESET times out`** (renamed in round 1)
   - Setup: a normal set, with `stopWorkoutBlock = awaitCancellation()`.
   - Asserts the RESET was attempted once and never completed.
   - Asserts disconnect ran once, at least 5000 ms of virtual time after the clear.
   - Asserts the arm row is still present.

**Revert check.** I stashed only `MainViewModel.kt`. Both tests then failed ("RESET must be attempted…", with `stopWorkoutCallCount` 0), and the other 42 tests still passed.

**Targeted run:** `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest --tests '*MainViewModelTest*'`, 44 tests, 0 failed (47 after round 1).

**Full run:** `./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest --continue`

| Suite | Tests | Failed |
|---|---|---|
| shared | 3805 | 2 | 
| androidApp | 52 | 0 |

The 2 shared failures are the known Windows baseline: `IosComposeResourceContractTest` `xcodeResourceStagingFailsWhenProcessedResourcesAreMissing` and `xcodeStagesResourcesForSimulatorAndDeviceTargets`.

`gradle.properties` was checked out, the daemon JVM file was removed, and the worktree is clean.

## Deviations from the plan
- **Test support edit outside the plan's file list.** The plan's list did not include `FakeBleRepository.kt`. The one-line `Event.Disconnected` is needed to observe that stop happens before disconnect: `disconnectCallCount` alone can't show order.
- **No edits to `DefaultWorkoutSessionManager.kt` or `ActiveSessionEngine.kt`.** They were not needed.
- **The predicate is narrower than the plan's** (added in round 1, R-10). The plan's check is purely state-based; the implementation also requires the link to be `Connected`, because off-link a RESET can only fail and waiting for it would defer `cleanup()` by the GATT timeout. This is within the decision's "attempt RESET ... where possible" and is pinned by the off-link test.

## Uncompiled iOS changes
None. All changes are in commonMain, commonTest and androidHostTest. **The fix is Android-only in practice** (corrected in review round 1, R-6): on iOS `MainViewModel` is a Koin `single` (PlatformModule.ios.kt:99) resolved with `koin.get()`, is never put in a `ViewModelStore`, and so `onCleared()` never runs there. Android gets it through `koinActivityViewModel()` (AndroidAppHost.kt:64).

- **RESET for non-live states.** When the state is SetSummary or Resting with teardown Ready, the lease check is true, so a RESET is sent. This is harmless on an idle trainer, and it is what the plan specifies.
- **Process death is not covered.** If the OS kills the process without calling `onCleared`, no RESET is sent. Only the hazard row (and the relaunch warning) covers that case. This is the app-scoped engine question, which is out of scope under KD-8.
- **Clean RESET on this path leaves the row.** Even a confirmed clean RESET here keeps the arm row, by design. The user acknowledges it on the next launch through PR 1's recovery dialog.
- **iOS has no teardown RESET at all** (R-6). `onCleared()` never runs there, and `WorkoutServiceController` is the no-op one, so there is no foreground component to keep the process alive for a RESET either. On iOS only the #782 hazard row and the relaunch warning cover a mid-set app exit.
- **The user-visible disconnect control is still RESET-less** (R-7). `BleConnectionManager.disconnect()` is a bare `scope.launch { bleRepository.disconnect() }` and is reachable during an active set (ActiveWorkoutScreen.kt:423, EnhancedMainScreen.kt:415). It is the same hazard class as this PR, and a far more likely user action. Applying the same lease-held-then-raw-RESET check there is the natural next step.
- **The predicate ignores a stored hazard row** (R-3, R-16). A row restored at startup, or an unresolved row for the currently connected trainer after a dismissed loss, leaves `workoutState` Idle and teardown `Ready`, so no RESET is attempted even though the app is connected and A-002 says to assume load. Widening the predicate to "an unresolved row for `transport.connectedTrainerAddress`" must keep this PR's rule: raw `bleRepository.stopWorkout()` only, never the engine teardown path, so the row is never resolved.
- **The tail can outlive the ViewModel** (R-2). With a lease, `cleanup()` and `disconnect()` now run up to the GATT timeout after `onCleared`, and both touch Koin singletons. If a successor ViewModel reconnected to the same trainer inside that window, the stale tail would stop its foreground service and drop its link. No guard was added, because a state-based skip cannot tell the successor's link from ours and a wrong skip leaks the radio. A real fix needs an ownership token on `WorkoutServiceController`/the BLE link.

### Production (`MainViewModel.onCleared`)
- **Log levels (R-1, R-9c, R-17, R-18).** A confirmed RESET logs at INFO. A failed `Result` or a timeout logs at `Logger.w` as "RESET before ViewModel teardown NOT confirmed", which is the line a field log is read for. The catch uses `Logger.e(e)` so the stack survives.
- **Skip the RESET off-link (R-2, R-10).** `holdsMachineLease` now also requires `connectionState is Connected`. Off-link the RESET can only fail, so this removes the whole up-to-5 s window — and the deferred foreground-service stop — for the not-connected case, including `RecoveryRequired` after a failed teardown on a dead link.
- **Comments (R-4, R-19, R-20).** Why cancellation is deliberately absorbed (the body is `NonCancellable`; rethrowing would skip `cleanup()` and leave the radio connected), that this body can outlive the ViewModel by the GATT timeout, and that the no-lease path still runs `cleanup()` synchronously on `Main.immediate`.
- **Readability (R-5).** The "MainViewModel cleared" log moved to the end of the coroutine, `workoutState.value` replaces the reach through the manager, and `BleConstants` / `withTimeoutOrNull` are imported.
- **No ownership guard (R-2, wontfix with justification).** A state-based skip cannot distinguish a successor that reconnected to the same trainer from our own link, and a wrong skip leaks the radio this block exists to close. Recorded as a follow-up instead.
- **Out of scope, recorded as follow-ups (R-3, R-7, R-16):** widening the predicate to a stored/visible hazard row, and the RESET-less `BleConnectionManager.disconnect()`.

### Tests (now 5; `MainViewModelTest` has 47)
- `clearing the ViewModel mid Just Lift set…` also pins `cleanup()`'s position (R-14) via `beforeCleanupGuardCloseForTest`: `StopWorkoutCompleted` < cleanup marker <= `Disconnected`.
- `clearing the ViewModel mid manual set still disconnects when RESET times out` (renamed, R-15) now does `runCurrent()` → `disconnectCallCount == 0` → `advanceUntilIdle()` → 1, and asserts `StopWorkoutEntered` before `Disconnected` (R-8, R-12). The virtual-time check is secondary.
- **New** `clearing the ViewModel mid set disconnects and keeps the arm row when RESET throws` (R-13).
- **New** `clearing an idle ViewModel sends no RESET and disconnects immediately` (R-11): no RESET, one disconnect, no virtual time elapsed.
- **New** `clearing the ViewModel after the link dropped mid set skips the RESET` (R-10).

### Mutation checks (each run on `*MainViewModelTest*`)

| Mutant | Failing tests |
|---|---|
| Whole production change reverted (file restored from `229fe6f2`) | the Just Lift, timeout and throwing tests (the two negative tests correctly still pass) |
| `&& connectionState is Connected` removed | exactly the off-link test |
| `cleanup()` moved above the RESET attempt | exactly the Just Lift test |

### Final run

| Suite | Tests | Failed |
|---|---|---|
| shared | 3808 | 2 |
| androidApp | 52 | 0 |

The 2 failures are the known Windows `IosComposeResourceContractTest` pair. `gradle.properties` was checked out, the daemon JVM file removed, and the worktree is clean.

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 5:04 AM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/815).

PR #820, opened 2026-09-20
Keep deleted workouts deleted, and scope "Delete All Workouts" to the current profile

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 4).

> **Stacked PR.** Based on `brownfield/62026a5c/pr-3-stop-pulls-from-rewriting-existing-sessi` (PR 3). Merge that first; GitHub retargets this PR automatically.

### PR 4: Keep deleted workouts deleted, and scope "Delete All Workouts" to the current profile
- Depends on: PR 3
- Files: shared/src/commonMain/sqldelight/com/devil/phoenixproject/database/migrations/48.sqm (new), shared/src/commonMain/sqldelight/com/devil/phoenixproject/database/PhoenixDatabase.sq, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/local/SchemaManifest.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/local/MigrationStatements.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/SqlDelightWorkoutRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/WorkoutRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/HistoryManager.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/SqlDelightSyncRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/sync/SyncManager.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/SettingsTab.kt, shared/src/commonMain/composeResources/values/strings.xml, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/data/repository/SqlDelightWorkoutRepositoryTest.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/data/repository/SqlDelightSyncRepositoryTest.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/presentation/manager/HistoryManagerTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/data/sync/SyncManagerTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/FakeWorkoutRepository.kt, shared/src/androidHostTest/kotlin/com/devil/phoenixproject/data/local/MigrationStatementsMirrorTest.kt (new)
- Traces to: FP-3, FP-4, FP-6, F-004, F-016, F-027
- Risk: medium
- Description: Implement KD-4.
  1. Add `48.sqm`. It creates `DeletedWorkoutSession(id TEXT PRIMARY KEY, portal_id TEXT NOT NULL, profile_id TEXT NOT NULL, deleted_at INTEGER NOT NULL)`, where `portal_id` = `routineSessionId ?: id`, the id the portal knows (R-19). Add indexes on `profile_id` and `portal_id`.
     - The same migration creates `PulledWorkoutSession(id TEXT PRIMARY KEY)` (no FK), the origin marker for R-10.
     - It backfills that table only with existing sessions that have **`updatedAt IS NOT NULL`** (already stamped by a sync or pull) **and** no MetricSample, RepMetric or CompletedSet rows (R-23). A never-uploaded local session (`updatedAt IS NULL`) is never marked, even when childless. Split-transaction saves (F-012) can leave such sessions, and marking them would stop them from ever being uploaded. Sessions the pull created have none of these; locally recorded ones do. Excluding a childless local session from later pushes can't destroy portal data, because its portal copy has no children from this phone either.
     - The table has **no** FK to `UserProfile`. Tombstones must outlive a permanently deleted profile (PR 20).
     - Mirror the table in `.sq` and add a `SchemaManifest` table op.
  2. `deleteSession`, `deleteSessionsByRoutineSessionId` and the new `deleteAllSessionsForProfile(profileId)` insert tombstones and hard-delete, in one transaction. `deleteAllSessionsForProfile` replaces the unscoped `deleteAllSessions`/`deleteAllWorkouts` for the user action.
  3. **Pull and push by origin** (R-10).
     - `mergeSessionsLww` records every id it inserts in `PulledWorkoutSession`, in the same transaction.
     - The session push gather (`getWorkoutSessionsModifiedSince`) excludes ids in `PulledWorkoutSession` unless the row has at least one local RepMetric or CompletedSet row.
  4. `mergeSessionsLww` also skips any pulled id present in `DeletedWorkoutSession`. The skip is **by id only and never profile-filtered**, so it still holds after the portal re-scopes a deleted profile's rows to Default (R-11).
  5. `SyncManager` adds the profile's tombstone `portal_id`s to the pull's `knownEntityIds` session list.
  6. The Settings dialog names the profile, for example "Delete all workouts for <profile>". Keep any other unscoped wipe (the full-data reset path) unchanged, and document which caller uses which.
  7. Remove the stale `.sq` comment at ":714-716".
  8. **Resilient-migration mirror** (R-15). Add the missing `47` entry and the new `48` entry to `getMigrationStatements` in `MigrationStatements.kt`. The resilient fallback in `DriverFactory.android.kt:48-67` and iOS `ResilientMigratingSchema` replays these statements when `Schema.migrate` throws, and entry 47 currently falls through to `else -> emptyList()`. Add `MigrationStatementsMirrorTest`. For every `N.sqm` it asserts that `getMigrationStatements(N)` is non-empty and matches the `.sqm` statement for statement. The test guards PRs 5, 11 and 23 as well.
- Acceptance:
  - New FK-on tests:
    - delete a session, pull the same id back: it stays gone;
    - a grouped routine workout's tombstone contributes its `portal_id` to the known ids;
    - a pulled session whose `updatedAt` is later bumped (a direct `UPDATE`) is still absent from the next push payload;
    - the 48 backfill marks a childless session with `updatedAt` set as pulled-origin. It leaves unmarked both a session with MetricSample/CompletedSet rows and a childless session with `updatedAt IS NULL`, and that last session is still pushed on the next sync;
    - Delete All under profile A leaves profile B's sessions and children intact.

    Each fails with this PR's production change reverted on its base branch (with PR 2's FK-on factory).
  - `MigrationStatementsMirrorTest` passes and fails when entry 48 is removed.
  - `SchemaParityTest` passes (fresh equals upgraded, including 48). The generated `Schema.version` is 49.
  - Full command: 0 failures. Revert Gradle dirt.

## Implementation summary
# PR 4 summary: deleted workouts stay deleted, pulled sessions are never re-pushed, Delete All is per profile

Branch: `brownfield/62026a5c/pr-4-keep-deleted-workouts-deleted-and-scope` (worktree D:/pwt/pr-4), stacked on PR 3 (98aebee7).
This PR's commits: **35fdb940**, **904b0391**, **4090006f**, **0cec4d88** and **0847bb86** (review round 1). Final SHA: **0847bb86**

### 1. Migration 48 (`48.sqm`, new) — two tables, both without a UserProfile FK
- `DeletedWorkoutSession(id PK, portal_id, profile_id, deleted_at)` plus `idx_deleted_session_profile` and `idx_deleted_session_portal`. `portal_id` = `COALESCE(routineSessionId, id)` — the id the portal knows (R-19).
- `PulledWorkoutSession(id PK)` — the pulled-origin marker (R-10).
- Backfill: `INSERT OR IGNORE INTO PulledWorkoutSession SELECT id FROM WorkoutSession WHERE updatedAt IS NOT NULL AND NOT EXISTS(MetricSample/RepMetric/CompletedSet)` (R-23). A childless row with `updatedAt IS NULL` is never marked; it can be a split-transaction save (F-012) that still owes an upload.
- Every statement is replay-safe (`CREATE TABLE/INDEX IF NOT EXISTS`, `INSERT OR IGNORE`), because `applyMigrationResilient` replays entry 48 after a partial migrate and only classifies "duplicate column"/"already exists" as recoverable — a bare `INSERT` would abort on `UNIQUE constraint failed`.
- Mirrored in `PhoenixDatabase.sq` (new section at the end, next to SessionNotes), in `SchemaManifest` (2 table ops + 2 index ops) and in `MigrationStatements` (entry 48, plus the missing entry 47).

### 2. Deletes write tombstones (`PhoenixDatabase.sq`, `SqlDelightWorkoutRepository`, `WorkoutRepository`)
- `deleteSession`, `deleteSessionsByRoutineSessionId` and the new `deleteAllSessionsForProfile(profileId)` each run `insertSessionTombstone*` + the DELETE in one `db.transaction`. The tombstone is a `SELECT` off the rows that are about to go, so it has to be written while they still exist.
- `deleteAllSessions` (unscoped) and its `.sq` query are **deleted**. It had exactly one caller, `HistoryManager.deleteAllWorkouts`; there is no other full-data wipe path in the codebase (grep for `deleteAllSessions` / `DELETE FROM WorkoutSession`), so nothing had to be "kept unchanged".
- `HistoryManager.deleteAllWorkouts()` deletes the active profile's history only (A-009 / F-016), falling back to `"default"` exactly like every read path in that class and like the profile name the dialog shows. (Round 1 changed this: it first refused the fallback and silently did nothing, which let the user confirm "Delete all workouts for Default?" over the default profile's history and then delete nothing.)

### 3. Pull and push by origin (`SqlDelightSyncRepository`, `PhoenixDatabase.sq`)
- `insertSessionIfAbsent` (the helper all three pull paths share — `mergeSessionsLww`, `mergePortalSessions`, `mergeAllPullData`) now: skips an id that is present locally **or** tombstoned (one `selectSessionKnownOrTombstoned` query), marks a real insert in `PulledWorkoutSession`, and returns whether it inserted. This covers the two insert paths named in the carryover.
- The tombstone skip is **by id only, never profile-filtered**, so it still holds after the portal re-scopes a deleted profile's rows to Default (R-11).
- `mergeSessionsLww` only attempts `updatePulledSessionTag` when the row was NOT inserted (a fresh insert already carries the incoming tag).
- `selectSessionsModifiedSince` (the push gather) excludes ids in `PulledWorkoutSession` unless the row has a local `MetricSample`, `RepMetric` or `CompletedSet` — one "this device has local data" predicate shared with the 48 backfill and `updatePulledSessionTag`. Exclusion is by origin, not timestamp, so a bumped `updatedAt` does not leak a pulled row back into the push (closes PR 3's R-9 duplicate-portal-workout risk).
- `SyncManager` adds `getDeletedSessionPortalIds()` to the pull's known session ids. It is **not** profile-scoped (matching the id-only merge skip, so a merged or re-scoped profile's deletions keep suppressing), a grouped workout contributes its `portal_id` only once no live row of the group is left, and tombstones get their own bounded slice of the parity budget (`SyncConfig.MAX_TOMBSTONE_PARITY_IDS = 500`, newest deletions first) so they can never evict live session ids at the cap.

### 4. Settings dialog names the profile (`SettingsTab`, `NavGraph`, `MainViewModel`, `strings.xml`)
- `delete_all_workouts_title` = "Delete all workouts for %1$s?", `delete_all_workouts_message` = "…permanently delete %1$s's workout history. Other profiles keep theirs…". Key names unchanged; no translation file carried these keys.
- `SettingsTab` takes a new `activeProfileName: String`; `MainViewModel.activeProfileName` is a `StateFlow<String>` next to `activeProfileId`.

### 5. Stale comments removed
- The `.sq` ":714-716" comment ("workout-session tombstone push is not currently implemented") and the same claim in `WorkoutRepository.kt`'s KDoc are gone; `softDeleteSessionsByRoutineSessionId`'s warning now says what user deletes do instead. `HistoryManager.deleteRoutineWorkouts`'s KDoc said "soft-delete" although it hard-deletes — fixed.

### 6. `ProfileDeletionMergePolicy.directProfileOwnedTables` (second commit)
`DeletedWorkoutSession` has a `profile_id`, and `SqlDelightUserProfileRepositoryTest.deletionPolicyCoversEveryDirectProfileOwnedSqliteTable` asserts that inventory matches the schema exactly. Added with a comment stating the policy for PR 20: a permanent profile delete must leave the tombstones untouched, a merge should reassign them to the target profile.

### 7. Review round 1 (commit 0cec4d88)
- **`discardSession`** (new `WorkoutRepository` member): the same hard delete with no tombstone, for rows this device wrote and never meant to keep. The assessment-save compensating rollback and the debug QA seeder use it, so a failed save or a fixture cleanup no longer blackballs an id forever.
- **Post-open backfill repair**: `MigrationManager.backfillPulledSessionMarkers()` replays migration 48's `INSERT OR IGNORE` (`backfillPulledWorkoutSessions`) behind a one-shot settings key inside `runRequiredMigrations`, so a database whose migration could not finish the backfill still gets its origin markers instead of silently reopening the R-10 re-push hole. A failure logs and leaves the key unset so the next open retries. This covers Android; on iOS a non-recoverable statement makes `ResilientMigratingSchema` fail the database open, which no post-open step can repair. The backfill is the only statement in 48 that can be non-recoverable, and only if `MetricSample`/`RepMetric`/`CompletedSet` is missing at migration time — `reconcileFullSchema` on the previous open makes that close to unreachable, and it is the same residual the design accepts for PR 5's migration 49. The backfill stays inside `48.sqm` because the plan (R-23) and `SchemaParityTest`'s acceptance test require it there.
- **Delete All fallback** and the **one local-data predicate**, described above.
- **`MIRROR_EXEMPTIONS`** replaces the flat exemption set: a reason per migration, asserted `mirror ⊇ file` for the three superset cases (11, 19, 31), and a test that nothing after migration 31 can be exempted.

## Tests

New:
- `MigrationStatementsMirrorTest` (new file, 3 tests): every `.sqm` with SQL has a non-empty mirror entry; mirror entries match their file statement for statement; there is one `.sqm` per schema step `1..Schema.version-1`.
- `SchemaParityTest.migration 47 to 48 marks only stamped childless sessions as pulled origin`: three seeded rows (stamped+childless, stamped+children, childless+unstamped); asserts only the first is marked and that the unstamped row is still returned by `selectSessionsModifiedSince` while the marked one is not. `EXPECTED_SCHEMA_VERSION` 48 → 49.
- `SqlDelightSyncRepositoryTest`: `a workout deleted on this device is not resurrected by a later pull` (also pulls the same id under a different profile id); `the push gather skips pulled sessions by origin, not by stamp` (pulled row with bumped `updatedAt` excluded, pulled row with a local `CompletedSet` and a locally captured row included); `newer pull does not re-tag a childless row this device captured`.
- `SqlDelightWorkoutRepositoryTest`: `deleteAllSessionsForProfile removes only that profile's sessions and children`; `deleting workouts tombstones them under the id the portal knows` (grouped → routineSessionId, standalone → id).
- `HistoryManagerTest.delete all workouts only deletes the active profile's history` (both halves) and `…falls back to the default profile like the history it shows`.
- `SyncManagerTest.pullSendsTombstonePortalIdsAsKnownSessionIds`.

Added in review round 1:
- `SqlDelightSyncRepositoryTest.a workout deleted on this device is not resurrected by the no-updatedAt pull path` (the `mergeAllPullData` branch: row stays absent, `PulledWorkoutSession` stays empty) and `tombstone portal ids are fed to the pull across profiles and only for dead groups`.
- `SqlDelightWorkoutRepositoryTest.discardSession drops the row without tombstoning the id`; the grouped-delete test now also asserts that a half-deleted group contributes no portal id.
- `MigrationManagerTest.startup backfills missing pulled-session markers exactly once`.
- `SyncManagerTest.pullKeepsTombstoneIdsWithoutEvictingLiveSessionIdsAtTheParityCap` (seeds `MAX_PARITY_IDS + 5` live ids plus one tombstone).
- `MigrationStatementsMirrorTest`: `superset exemptions still replay every statement of their migration file`, `no migration after 31 is exempt…`, `no migration file hides a statement separator inside a string literal`.

Updated: `SqlDelightWorkoutRepositoryTest.deleteAllSessions removes all sessions` replaced by the scoped test; `ProfileSettingsSeparationContractTest` pinned signature + "Settings collects exactly N flows" 3 → 4 + the new `activeProfileName` wiring line; `FakeWorkoutRepository` and `FakeSyncRepository` follow the interfaces.

### Revert check
All production behaviour was neutered on top of commit 35fdb940 (tombstone writes removed, tombstone skip disabled, `markSessionPulled` removed, the `PulledWorkoutSession` clauses removed from `selectSessionsModifiedSince` and `updatePulledSessionTag`, `deleteAllSessionsForProfile` made unscoped in SQL and in the fake, `getDeletedSessionPortalIds` dropped from the known ids, mirror entry 48 renamed away, the 48 backfill deleted). Result: **10 failures**, including every new test except one confound —
- MirrorTest (both assertions), SchemaParityTest 47→48, `the push gather skips pulled sessions by origin`, `a workout deleted on this device is not resurrected`, both workout-repository tests, `pullSendsTombstonePortalIdsAsKnownSessionIds`, `delete all workouts only deletes the active profile's history`, plus PR 3's `newer pull carries an exercise re-tag…` (the disabled skip-check also short-circuits the re-tag call).
- `newer pull does not re-tag a childless row this device captured` passed in that combined run for the same confound, so it was re-checked on its own: removing ONLY the `AND EXISTS (… PulledWorkoutSession …)` clause makes it fail, and PR 3's two child-guard tests still pass.

No second mutation pass was run for round 1; each new test pins one clause by construction: `discardSession drops the row without tombstoning the id` fails if `discardSession` writes a tombstone (i.e. if it delegates to `deleteSession`); `startup backfills missing pulled-session markers exactly once` fails without the `MigrationManager` call and again without the one-shot key; `tombstone portal ids are fed to the pull across profiles and only for dead groups` fails without either the `NOT EXISTS (live row)` clause or the removal of the profile filter; `pullKeepsTombstoneIdsWithoutEvictingLiveSessionIdsAtTheParityCap` fails if the tombstone budget is dropped (the tombstone falls off the tail); the `MetricSample` row in the push-gather test fails without that `OR EXISTS`; and the mirror test's three new cases fail if the exemption map loses its superset assertion, its `> 31` guard or the literal check. One exception: `a workout deleted on this device is not resurrected by the no-updatedAt pull path` is coverage of the shared insert helper on a second live caller, not a distinct production clause — the same guard the `mergeSessionsLww` test already pins.

### Full run (after review round 1)
`./gradlew -Pskip.supabase.check=true :shared:testAndroidHostTest :androidApp:testDebugUnitTest :shared:validateSchemaManifest :shared:verifySqlDelightMigration --continue`
- exit 0 / BUILD SUCCESSFUL (run on the tree of the final commit 0847bb86)
- shared: 3807 tests, 0 failed, 2 skipped (the two `IosComposeResourceContractTest` xcode cases)
- app: 52 tests, 0 failed
- `:shared:validateSchemaManifest` — "405 columns across 50 tables, all covered"
- `:shared:verifySqlDelightMigration` (aggregates `verifyCommonMainPhoenixDatabaseMigration`) — passes, so the `.sqm` chain and the `.sq` CREATE statements agree. My earlier claim that this task does not exist was wrong (R-18); it is worth running in PRs 5, 11 and 23 too.
- The pre-review run on 4090006f was 3799 shared / 52 app, also 0 failures.
- Working tree clean; `gradle.properties` untouched and no `gradle/gradle-daemon-jvm.properties`. No iOS-only code was changed. The only `shared/build.gradle.kts` edit is a comment: `Schema.version` is derived from the `.sqm` files (now 49) and the `version = 47` literal there is inert.

## Deviations from the plan
- **Carryover line 14 ("gate `updatePulledSessionTag` on the marker *instead of* the no-children heuristic") is implemented as marker AND no-children** (children now meaning MetricSample, RepMetric or CompletedSet). A backup restore can write `MetricSample`/`RepMetric`/`CompletedSet` rows onto a row that was originally pulled (and is therefore marked), and those children are local measurements — the marker alone would let a portal tag overwrite the tag of a row we now have data for. Keeping both gates is strictly more conservative and keeps PR 3's two child-guard tests meaningful. The marker clause is independently proven load-bearing (see the revert check).
- `deleteAllSessions` was deleted outright rather than "kept for the full-data reset path" (plan step 6): that path does not exist in this codebase.
- Four files outside the plan's list were touched because the suite/acceptance requires it: `ProfileSettingsSeparationContractTest.kt` (pinned SettingsTab signature, flow count 3 → 4), `ProfileDeletionMergePolicy.kt` (profile-owned table inventory), `SchemaParityTest.kt` (schema version + the migration-48 acceptance test) and `FakeSyncRepository.kt` (new interface member).
- `MigrationStatementsMirrorTest` cannot assert an exact mirror for **all** migrations: 3, 4, 5, 6, 8, 9, 10, 11, 19 and 31 diverge, and for four different reasons. `MIRROR_EXEMPTIONS` records one per migration — rebuild (3, 4, 8, 10), heal-covered (5: the mirror omits 5.sqm's four `UserProfile` auth `ALTER`s, which the `SchemaManifest` heal ops add on every open), forward-shape `CREATE TABLE`s (6), documented under-cover (9: the mirror skips the superset composite-id data step), and mirror-superset (11, 19, 31, asserted as `mirror ⊇ file`). Everything else (12–18, 20–30, 32–48) is compared statement for statement, and a test fails if any migration after 31 is added to the map, so PRs 5, 11 and 23 are guarded mechanically.
- Round 1 added eight more files outside the plan's list, all required by the fixes: `MigrationManager.kt` and `MigrationManagerTest.kt` (post-open backfill), `DataBackupManager.kt` (a note where the post-restore backfill would have gone), `SqlDelightAssessmentRepository.kt` and `SqlDelightAssessmentRepositoryTest.kt`, `ProfileQaSeeder.kt` and `ProfileQaSeederTest.kt` (`discardSession`), and `shared/build.gradle.kts` (stale comment). Fourteen files in total are outside the plan's list.

- **PR 8 owns the group-complete push (review R-19, design.md PR 8 step 0).** The portal's `replace_session_children` deletes and re-inserts a workout's children from whatever the payload carries, so a push that contains only some rows of a grouped workout drops the others' portal-side data. Origin exclusion makes such a partial group possible for good rather than self-healing, which is why PR 8's "send every row of a routine session together" is now load-bearing. Not fixed here: changing the gather's grouping semantics is outside a PR about origin marking, and the same partial-group delete exists pre-PR via the timestamp filter.
- `deleteSession(id)` on ONE row of a grouped workout tombstones that row under the whole workout's `portal_id`. The known-ids half of that is now handled (a group's id is only sent once no live row is left), but the merge-side skip stays id-only, so the deleted row itself can never be re-pulled. That is KD-4/R-19 semantics as designed; PR 10/PR 20 should be aware of it.
- Tombstones are never pruned and `PulledWorkoutSession` rows survive their session's deletion. Both tables are tiny (one TEXT id per row) and tombstones now have a bounded slice of the pull request, so the cost is capped; PR 20 / a portal `deletedSessionIds` channel (open question 3 in the design) is where a prune would belong — a tombstone must not be dropped before the portal knows about the deletion.
- The push still does not tell the portal about deleted workouts; the table is now ready to feed that channel.
- **Backup/restore (PR 21/22) still does not carry either new table, and the obvious interim fix does not work.** I implemented the suggested "re-run the origin backfill at the end of an import" and the round-trip test failed: restore inserts sessions through `insertSession` **without** `updatedAt` (badges, stats and notes carry their sync fields; sessions do not), so every restored session has a null stamp and the heuristic marks none of them. The hook was removed rather than shipped as a no-op and `DataBackupManager` carries a note. PR 22 must export/import `DeletedWorkoutSession` and `PulledWorkoutSession` (or preserve the session stamp). Consequence today: restored sessions are pushed as ordinary unstamped local rows, so the restore hazard is the portal's `replace_session_children`, i.e. PR 8's group guard, not the marker.
- PR 20 should reassign `DeletedWorkoutSession.profile_id` on a profile **merge** and leave it alone on a permanent **delete**; the `ProfileDeletionMergePolicy` comment says so. Nothing breaks if it does neither, because the known-ids feed is no longer profile-scoped.
- `shared/build.gradle.kts` still declares `version = 47`. It is inert (`Schema.version` is derived from the `.sqm` files and is 49); only the misleading comment was fixed, since changing the literal is not this PR's risk to take.
- `mergePortalSessions` and `PortalPullAdapter.toWorkoutSessions` still have no production callers (carried over from PR 3).

## Suggested carryover lines (not written to carryover.md)
- PR 8: step 0's group-complete push is now load-bearing, not just an improvement (review R-19). Origin exclusion is permanent, so a group that mixes a marked row with a live sibling would have the marked row's exercise deleted from the portal by `replace_session_children` on the next push of any sibling. The guard in design.md:588 is already written against PR 4's `PulledWorkoutSession`; note that "has local data" now means MetricSample OR RepMetric OR CompletedSet.
- PR 10: migration 48's backfill marks any stamped childless session, including locally recorded rows that were merely pushed once (review R-4). Those rows are permanently push-excluded, so a watermark reset or a forced full re-push will not re-upload them.
- PR 20: on a profile **merge**, reassign the tombstones (`UPDATE DeletedWorkoutSession SET profile_id = <target> WHERE profile_id = <source>`); on a **permanent delete**, leave them untouched — they are what stops the portal's re-scoped rows from reappearing under Default. `ProfileDeletionMergePolicy`'s comment says the same.
- PR 22: export and import `DeletedWorkoutSession` and `PulledWorkoutSession`. Restore also drops a session's `updatedAt`, which is why re-running the marker backfill after an import marks nothing (review R-24) — preserving the stamp is the alternative fix.
- PR 5 / 11 / 23: run `:shared:verifySqlDelightMigration` alongside the test suite; it verifies `.sqm` against the `.sq` CREATE statements directly. `MigrationStatementsMirrorTest` now fails if a migration after 31 is added to `MIRROR_EXEMPTIONS`, so each new migration must mirror its `.sqm` statement for statement in `MigrationStatements.kt`.

## Review


9thLevelSoftware:
### Merge activity

* **Sep 20, 12:09 PM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/820).

PR #821, opened 2026-09-20
Bound every machine command by the connected model's ceiling and a single progression limit; delete unvalidated builders

9thLevelSoftware:
Part of the brownfield audit/improvement run `62026a5c` (plan PR 13).

> **Stacked PR.** Based on `brownfield/62026a5c/pr-7-attempt-reset-before-disconnect-when-the` and also merges `brownfield/62026a5c/pr-30-correct-the-agent-facing-docs-claude-md`. The diff against its base includes those merged branches; this PR's own commits are listed in the implementer summary below.

### PR 13: Bound every machine command by the connected model's ceiling and a single progression limit; delete unvalidated builders
- Depends on: PR 7, PR 30
- Files: shared/src/commonMain/kotlin/com/devil/phoenixproject/util/WorkoutCommandValidator.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/util/BlePacketFactory.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/util/Constants.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/util/HardwareDetection.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/util/CommandLimits.kt (new), shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/BleRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/repository/KableBleRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/data/preferences/PreferencesManager.kt, shared/src/iosSimulatorArm64Main/kotlin/com/devil/phoenixproject/data/repository/PhantomBleRepository.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/ActiveSessionEngine.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/RoutineFlowManager.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/manager/DefaultWorkoutSessionManager.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/SetReadyScreen.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/RestTimerCard.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/RoutineOverviewScreen.kt, shared/src/commonMain/kotlin/com/devil/phoenixproject/presentation/screen/ExerciseEditBottomSheet.kt, shared/src/commonMain/composeResources/values/strings.xml, shared/src/commonTest/kotlin/com/devil/phoenixproject/util/WorkoutCommandValidatorTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/util/BlePacketFactoryTest.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/util/CommandLimitsTest.kt (new), shared/src/commonTest/kotlin/com/devil/phoenixproject/testutil/FakeBleRepository.kt, shared/src/commonTest/kotlin/com/devil/phoenixproject/presentation/manager/StoredRoutineLimitsTest.kt (new)
- Traces to: FP-1, F-009, F-010, F-020, F-044, F-050, F-059, A-001, A-011
- Risk: medium
- Description: Implement KD-9. The rule is **clamp at resolution, reject only as a backstop**, so existing routines keep starting (R-2, R-17).
  1. **Expose the model.** `BleRepository` exposes the connected `PhoenixModel` (it already exists at `KableBleConnectionManager.kt:676`). Persist the last connected model in `PreferencesManager`. Add `maxWeightPerCableKg(model)`: V-Form 100, Trainer+ 110, Unknown/null 100.
  2. **One shared clamp.** New `CommandLimits.resolve(weightKg, progressionKg, model)` replaces both `clampUpcomingProgressionKg` copies (`ActiveSessionEngine.kt:7228` with its four call sites, and `RoutineFlowManager.kt:174` with its call sites). Every parameter builder calls it **at command-resolution time**: set 1, later sets, recovery replay (`ActiveSessionEngine.kt:7605`), DWSM next-exercise (`DefaultWorkoutSessionManager.kt:1216`) and Just Lift.
     - It clamps progression to ±3.0 kg.
     - It clamps weight, in kg **after** unit conversion, to the connected model's ceiling.
     - It returns what it changed. The engine shows a one-time-per-set notice, for example "Weight capped to 100 kg/cable for this trainer" or "Progression capped to 3 kg/rep".
     - Stored routine values are not rewritten, so this also covers portal pulls, backups and CSV imports.
  3. **Validator as backstop.** `WorkoutCommandValidator.validateProgramParams` and the Echo checks take the ceiling. They reject:
     - `weightPerCableKg > ceiling + 0.05` (a tolerance for float and lb conversion, e.g. 220.5 lb ≈ 100.02 kg);
     - `|progressionRegressionKg| > 3.0`;
     - non-finite values.

     Validate `eccentricPct` before `createEchoControl` clamps it (F-010). A rejection at start shows a user message instead of a silent Idle. It is only reachable by bypassing step 2.
  4. `BlePacketFactory` builders take the ceiling, and the engine passes the connected model.
  5. **Sliders and live adjust.**
     - Live `adjustWeight` (`:6595`) and the SetReady/rest sliders use the connected model's ceiling.
     - Planning/editor sliders (`RoutineOverviewScreen`, `ExerciseEditBottomSheet`) use the last connected model, or 110 when none is known, so Trainer+ owners can plan offline.
     - The editor progression slider range becomes ±3 kg in the display unit. `ExerciseConfigViewModel` save-coercion is done in PR 5 (R-7).
  6. **Delete unvalidated builders.** Remove `KableBleRepository.startWorkout`, `sendInitSequence` and their interface members, the Phantom overrides, and `ActiveSessionEngine.sendWeightUpdateToMachine`.
  7. **Frame-asserting fake.** `FakeBleRepository.sendWorkoutCommand` decodes 96-byte program and 32-byte Echo frames and throws `AssertionError` on a weight above the fake's model ceiling or a progression above ±3. The model defaults to Trainer+, and tests set V-Form where relevant.
- Acceptance:
  - Table tests:
    - `CommandLimits`: progression 5 → 3; V-Form 105 → 100; Trainer+ 105 → 105; Unknown 105 → 100; lb 220.5 on V-Form → accepted.
    - Validator: V-Form 100.04 accepted, 100.1 rejected; Trainer+ 110.1 rejected; progression ±3.1 and ±50 rejected.
  - `StoredRoutineLimitsTest`: a stored routine with progression 5 and a V-Form routine at 105 kg/cable **both start**. The fake frames carry 3 and 100, and the capped notice is emitted.
  - Set 1 and set 2 of a routine send the same progression.
  - The whole engine suite passes with the frame-asserting fake, with arming ON (PR 1 harness).
  - Each new test fails with this PR's production change reverted.
  - Grep:
    - no `clampUpcomingProgressionKg`;
    - no `fun startWorkout(params` in `BleRepository`.
  - The CI iOS compile job is green. The reviewer checks `PhantomBleRepository` by hand, since CI doesn't compile it.
  - Full command: 0 failures beyond the baseline. Revert Gradle dirt.

## Implementation summary
# PR 13 summary: bound every machine command by the connected model's ceiling and a single progression limit

Branch: `brownfield/62026a5c/pr-13-bound-every-machine-command-by-the-conne` (worktree `D:/pwt/pr-13`).
Stacked on PR 7 (`6567d9f3`) → PR 6 → PR 1, then merged with PR 30's docs branch.

Commits for this PR:
- `57d3ba39` — merge of `brownfield/62026a5c/pr-30-correct-the-agent-facing-docs-claude-md` (clean, no conflicts)
- `969ff81b` — the PR itself
- `fa685acc` — self-review follow-ups (notice timing, log label, two tests)
- `2fe5b5d5` — review round 1 (all 43 items; per-item Status/Response in `reviews/pr-13.md`)
- `5a289c64` — round-1 self-review: the capped notice is shown on a scope that survives draining it

## What changed and why (KD-9; F-009, F-010, F-020, F-044, F-050, F-059)

The rule is **clamp at command resolution, reject only as a backstop**, so an existing routine keeps
starting instead of being stranded mid-workout (R-2, R-17). Firmware caps force and per-rep
progression (A-001/A-011), so this is defense in depth.

### 1. `util/CommandLimits.kt` (new) — the one bound
- `maxWeightPerCableKg(model)`: V-Form 100, Trainer+ 110, Unknown/null **fails closed to 100**.
- `planningMaxWeightPerCableKg(model)`: V-Form 100, everything else 110. Planning/editor screens run
  offline and command nothing, so an unknown model must not narrow a Trainer+ owner's planning range.
- `MAX_PROGRESSION_KG = 3.0`, `clampProgressionKg()` — the single definition that replaces both
  former `clampUpcomingProgressionKg` copies.
- `WEIGHT_TOLERANCE_KG = 0.05` — one constant used by three consumers: the "was this capped?" flag,
  the validator, and the frame-asserting fake. 220.5 lb is 100.017 kg, so a lb user at the machine
  maximum is neither rejected nor shown a spurious notice.
- `resolve(weightKg, progressionKg, model)` returns the bounded values plus `weightCapped` /
  `progressionCapped`. **Non-finite values are passed through untouched** so the validator rejects
  them rather than a clamp inventing a plausible number.

### 2. The model reaches the command path
- `BleRepository.connectedModel` is a **default interface property** reading
  `(connectionState.value as? Connected)?.hardwareModel`. The model was already detected at connect
  time (`KableBleConnectionManager.kt:676`); nothing needed a new override.
- `UserPreferences.lastConnectedModel` + `PreferencesManager.setLastConnectedModel` (key
  `last_connected_model`). `MainViewModel` collects `connectionState` and stores it, **skipping
  Unknown** so a bad name read can't narrow a known Trainer+ owner's planning sliders.

### 3. One clamp, at the single send site
Every start path — set 1, later sets, recovery replay, DWSM's next-exercise advance and Just Lift —
converges on `ActiveSessionEngine` ~8454: `bleParams` → validate → `createProgramParams` →
`sendWorkoutCommand`. `CommandLimits.resolve` runs there and produces a separate `commandParams`
used for validation and the frame. **`bleParams` itself is left alone**, because
`retryRackCommandMatches` (`:8404`, `:8556`) compares it against the captured retry request; clamping
it in place would fail-close a V-Form user's drop-set retry.

Because the clamp is at resolution, it also covers values that arrived from a portal pull, a backup
or a CSV import **without rewriting stored data** — which is why no `DataBackupManager` change was
needed (carryover note about PR 21's streaming insert sites therefore does not apply here).

`emitCommandLimitNotice` runs **right after `sendWorkoutCommand` succeeds**, not at resolution, so a
start refused by the #782 safety barrier, by the machine-configuration claim or by the validator
never announces a cap for a set that never began. It publishes to
`coordinator.commandLimitNotice`, a drainable `StateFlow` (round 1, R-1/R-36): the screen that
shows the notice is often not composed yet at send time, because Just Lift skips the countdown and
only navigates to `ActiveWorkoutScreen` once the state turns `Active` — an emission on the
replay-0 feedback flow was simply dropped. `ActiveWorkoutScreen` drains it on arrival, so it is
shown exactly once. An Echo command never announces a cap, because the 32-byte Echo frame carries
neither a weight nor a progression (R-5/R-17), and the wording says "(trainer model not
recognised)" rather than "for this trainer" when the model is Unknown (R-20). The Issue390 trace at
the resolution site is labelled `BLE PARAMS PRE-LIMIT`, because it logs the values before the
clamp.

### 4. Validator as the backstop
- `validateProgramParams(params, maxWeightPerCableKg)` — the ceiling is now **required**: the
  backstop should never guess. It rejects `weight > ceiling + 0.05`, `|progression| > 3.0`, and
  non-finite values.
- Two non-send callers pass `Constants.MAX_WEIGHT_PER_CABLE_KG` (the absolute hardware maximum) with
  an inline reason: `ActiveSessionEngine.validateWorkoutCommand` (a drop-set retry feasibility gate)
  and `DropSetCandidateResolver` (a candidate is always *lower* than the weight that just failed, so
  a model ceiling can never be the reason it is rejected there).
- `createEchoControl` now validates `eccentricPct` **before** clamping (F-010). The bound is
  reachable again: 200% is rejected instead of silently rewritten to 150%. `EccentricLoad` tops out
  at 150, so no production path is affected.
- `BlePacketFactory.createProgramParams`, `createWorkoutCommand` and `validateLegacyWorkoutCommand`
  **require** `maxWeightPerCableKg` too (round 1, R-3/R-16/R-33): a builder that guesses the ceiling
  reopens exactly the bypass this PR closes. The engine passes the connected model's ceiling; test
  call sites pass `CommandLimits.TRAINER_PLUS_MAX_WEIGHT_PER_CABLE_KG` through two local helpers.

### 5. Sliders
- `adjustWeight` (live): hardcoded `coerceIn(0f, 110f)` → the connected model's ceiling.
- `SetReadyScreen`, `RestTimerCard` (via `WorkoutTab`) and `JustLiftScreen` (round 1, R-34):
  connected model's ceiling. `SetReadyScreen` falls back to the planning ceiling while disconnected
  (round 1, R-42), so a Trainer+ owner is not narrowed to 100 before they connect.
- `RoutineOverviewScreen`, `ExerciseEditBottomSheet` (via `RoutineEditorScreen` and
  `SingleExerciseScreen`): `planningMaxWeightPerCableKg(userPreferences.lastConnectedModel)`.
- Editor progression slider: `maxWeightChange` is now `kgToDisplay(3 kg)` — ±3 in kg, ±6 in lb,
  rounded **down** so the slider can't offer a step the clamp would immediately cap.
  `ExerciseConfigViewModel` save-coercion stays in PR 5 (R-7).
- `WeightChangePerRepControl` derives its own ± range from `CommandLimits.MAX_PROGRESSION_KG`
  instead of two private constants (round 1, R-4/R-25), and is **display-only** for out-of-range
  input — see root cause (A) below.
- `SliderWithButtons` labels `value.coerceIn(valueRange)` (round 1, R-18), so a stored 105 on a
  100 kg slider no longer reads "105 kg" with the thumb pinned at the right edge.

### 6. Deletions
- `BleRepository.startWorkout(params)` and `sendInitSequence()` — removed from the interface and from
  `KableBleRepository`. Both hand-built command bytes with no validator call and had no production
  callers.
- `ActiveSessionEngine.sendWeightUpdateToMachine` — no callers; its own comment said the frame it
  built would fault the machine mid-set.
- Both `clampUpcomingProgressionKg` copies, and the pre-clamps at their call sites (see Deviations).

### 7. Frame-asserting fake (F-050)
`FakeBleRepository.sendWorkoutCommand` decodes 96-byte program frames (targetWeight 0x58,
progression 0x5C, forceMax 0x54, reps 0x04) and 32-byte Echo frames, and throws `AssertionError` on a
weight outside `0..ceiling+0.05` or `|progression| > 3`, and on an Echo frame whose eccentric
overload is outside 0..150 (round 1, R-24 — the field F-010 is about). `simulateConnect` resolves
`hardwareModel` from the device name exactly as production does (round 1, R-22), so every
`simulateConnect("Vee_Test")` in the suite genuinely runs at the V-Form 100 kg ceiling and the wider
bound has to be asked for. Unrecognised frame shapes are **denied by default** (round 1, R-23): only
a whitelist of zero-load control frames passes, so a revived load-carrying builder has to be taught
to the fake before the suite will accept it.
The dead `workoutParameters` list (only ever fed by the deleted `startWorkout`) was replaced by
`programCommands` / `echoCommands`, which record real decoded frames. This turned ~50 vacuously-true
assertions across 9 engine tests into real ones; exactly one of them was wrong (see Deviations).

## Tests

New:
- `commonTest/util/CommandLimitsTest.kt` — 6 tests: per-model ceilings incl. Unknown failing closed;
  the planning ceiling opening up; progression 5 → 3 and −50 → −3; V-Form 105 → 100, Trainer+ 105 →
  105, Unknown 105 → 100; **lb 220.5 on V-Form accepted with no capped notice**; non-finite passthrough.
- `commonTest/presentation/manager/StoredRoutineLimitsTest.kt` — 6 tests on the default harness
  (4 in round 0, 2 added in round 1)
  (arming ON): a stored routine with progression 5 **starts**, frame carries 3, notice emitted, stored
  routine unchanged; a 105 kg/cable routine **starts** on a V-Form, frame carries 100 (forceMax 110),
  notice emitted, stored routine unchanged; the same routine is commanded in full on a Trainer+ with
  no notice; set 1 and set 2 send the same progression `[3, 3]`.
- `WorkoutCommandValidatorTest` — 3 new tests: per-model ceiling (105 rejected on V-Form, accepted on
  Trainer+, rejected on Unknown); tolerance (100.04 accepted / 100.1 rejected on V-Form, 110.04 /
  110.1 on Trainer+); progression bound (0, ±3, 1.5 accepted; ±3.1, ±50 rejected; NaN and Inf rejected).
- `BlePacketFactoryTest` — 2 new tests (progression beyond the limit rejected at the frame; weight
  above the model ceiling rejected on V-Form and accepted on Trainer+), and the Echo 200% test now
  asserts rejection instead of clamping.
- `MainViewModelTest.connecting remembers the trainer model for the offline planning sliders` — the
  only new production path that had no coverage: a connect stores the model, and an Unknown model
  never overwrites a known one.

### Revert / mutation evidence (each mutant applied alone, then reverted)

| Mutant | Failing tests |
|---|---|
| `CommandLimits.resolve` + notice removed at the send site (`commandParams = bleParams`) | 3 of 4 `StoredRoutineLimitsTest`, `RestTimerProgressionWiringTest`, **and `DropSetCompletionCaptureTest`** — the frame-asserting fake caught the unbounded frame in an unrelated engine test, which is the F-050 invariant doing its job |
| `validateProgramParams` progression bound removed | `WorkoutCommandValidatorTest.program params bound per-rep progression`, `BlePacketFactoryTest.program frame rejects a per-rep progression beyond the command limit` |
| `createEchoControl` clamps before validating again | `BlePacketFactoryTest.Echo eccentric load above 150 percent is rejected, not clamped` |
| Fake stops decoding frames (`decodeAndAssertWithinLimits` removed) | all 4 `StoredRoutineLimitsTest` |

### Grep acceptance
- `clampUpcomingProgressionKg`: 0 hits in production code. The single remaining occurrence is
  the string literal inside `RestTimerProgressionWiringTest`, which asserts its absence.
- `fun startWorkout(params` in `BleRepository`: 0 hits.

### Final run

| Suite | Tests | Failed |
|---|---|---|
| shared | 3829 | 2 |
| androidApp | 52 | 0 |

The 2 failures are the known Windows baseline: `IosComposeResourceContractTest`
`xcodeResourceStagingFailsWhenProcessedResourcesAreMissing` and
`xcodeStagesResourcesForSimulatorAndDeviceTargets` (need `/bin/sh`).

No Gradle dirt: `gradle.properties` unmodified, no `gradle/gradle-daemon-jvm.properties`, worktree
clean at `5a289c64`.

## Deviations from the plan

- **`strings.xml` was not touched.** Every existing engine feedback message
  (`_userFeedbackEvents`, `_bleErrorEvents`) is a plain English literal in Kotlin, and localizing one
  of them alone would be inconsistent. (Corrected after R-39: the engine *could* resolve one —
  `org.jetbrains.compose.resources.getString` is a plain suspend function usable outside
  composition — so the reason is convention, not a technical block, and the eventual localization
  pass is scoped as "all engine feedback strings".) The capped notice follows that convention:
  `"Weight capped to 100 kg/cable for this trainer"` / `"Progression capped to 3 kg/rep"`. Localizing
  the engine's user-facing strings is a pre-existing gap, noted as a follow-up.
- **The notice is in kg regardless of the user's display unit.** The ceiling is a per-cable kg
  hardware figure and the existing engine strings do no unit conversion. Follow-up below.
- **The per-call-site progression pre-clamps were removed, not just repointed.** The plan says
  `resolve` "replaces both `clampUpcomingProgressionKg` copies … with its call sites". Initially I
  kept the call sites (routed through `CommandLimits.clampProgressionKg`), and the acceptance notice
  then never fired: `enterSetReady` had already clamped 5 → 3 before the command was resolved. The
  clamp now happens **only** at command resolution. `updateSetReadyProgressionKg` keeps
  `CommandLimits.clampProgressionKg` — it is direct user input, not a builder copy.

  **Corrected after review round 1 (R-2, R-12, R-13, R-29).** Round 0's version of this note said
  "Set Ready surfaces the stored value (5) and the command carries 3 with a notice; the
  `WeightChangePerRepControl` clamps its own display and syncs the clamped value back to the parent,
  so the on-screen control still shows a legal value." With the control rendered, neither half was
  true. Its `SideEffect` called `onValueChangeKg(3f)` — the **user-edit** handler — on first
  composition, which (a) wrote 3 into `_workoutParameters` before the user tapped Start, so
  `progressionCapped` was already false and no notice fired, and (b) on the rest card latched
  `_userAdjustedWeightDuringRest`, so the next advance inherited the previous set's weight and reps.
  Round 1 deletes that write-back: the control is display-only for out-of-range input,
  `CommandLimits.resolve` is the only thing that changes a commanded value, and the notice is
  reachable on every path.

  **Precisely what Set Ready shows now**, since the round-0 wording was loose in the other
  direction too: the *state* (`RoutineFlowState.SetReady.adjustedProgressionKg` and
  `_workoutParameters.progressionRegressionKg`) holds the stored 5, and that is what
  `CommandLimits.resolve` sees, so the notice fires. The *control* still renders the clamped
  +3 kg/rep, read-only — it simply never writes that 3 back. So the user sees a control at +3,
  taps Start, and then gets "Progression capped to 3 kg/rep". That reads a little oddly, and it is
  the agreed outcome rather than an oversight: read-only rendering is what R-12/R-29 asked for, and
  the alternative (rendering 5 on a slider whose range stops at 3) is worse. Showing the cap on the
  SetReady screen *before* Start would be the nicer UX and is listed as a follow-up.
- **PhantomBleRepository: `override` removed, methods kept.** `startWorkout`/`sendInitSequence` are no
  longer `BleRepository` members, so nothing can reach them through the interface. Deleting the bodies
  would have gutted the phantom simulator (`workoutParams` drives its metric/rep simulation) and ~20
  `PhantomBleRepositoryTest` call sites in code this host cannot compile. They are now
  simulator-local methods. **This file is uncompiled here and is not compiled by CI either** (CI runs
  `:shared:compileKotlinIosArm64`, not the simulator target) — the reviewer should read it by hand.
  `PhantomBleRepository` has no DI binding; it is test-only.
- **`MainViewModel.kt` is outside the plan's file list.** The `lastConnectedModel` writer has to sit
  where both `connectionState` and `PreferencesManager` are available, and the plan's listed files
  (`KableBleRepository`) have no preferences access. No new collector class was added — one
  `viewModelScope.launch` in the existing init block.
- **`DropSetCandidateResolver.kt`, `WorkoutTab.kt`, `RoutineEditorScreen.kt`,
  `SingleExerciseScreen.kt` are outside the plan's file list.** Each was forced by the required
  validator parameter or by passing the ceiling down to a listed composable.
- **Three existing tests were updated** because the behaviour they pinned changed deliberately:
  - `DWSMRoutineFlowTest.enterSetReady_clampsOutOfRangeRoutineProgressionToControlRange` → renamed to
    `enterSetReady_surfacesStoredProgressionUnchangedAndLeavesTheBoundToTheCommand`.
  - `RestTimerProgressionWiringTest.activeSessionEngine_preservesRestProgressionEditsWhenAdvancing` —
    the source-text assertion requiring `clampUpcomingProgressionKg(...)` now asserts its absence plus
    the presence of `CommandLimits.resolve` and `emitCommandLimitNotice`.
  - `DWSMWorkoutLifecycleTest.terminal timer expiry cannot skip after a concurrent extend and pause` —
    `assertEquals(starts, …)` became `starts + 1`. With the fake's real frame list this assertion is
    no longer vacuous, and the same test already asserts `currentSetIndex == 1`, i.e. autoplay
    legitimately started set 2. **This is a case where a previously always-true assertion was hiding
    a real observation; worth a reviewer's eye.**
- **`validateEchoControl` does not take the ceiling** (R-40), although plan step 3 says "the Echo
  checks take the ceiling". The 32-byte Echo control frame encodes neither a target weight nor a
  per-rep progression, so there is no load field for a ceiling to bound. The Echo frame does get its
  own bound: the fake asserts its eccentric overload is 0..150 (R-24), and `createEchoControl` now
  validates `eccentricPct` before clamping (F-010).
- **`Constants.kt` is in the plan's file list and was not changed** (R-38).
  `MAX_WEIGHT_PER_CABLE_KG` is still the right "absolute hardware maximum" constant and has
  non-command callers (`MonitorDataProcessor`, `ApplyEquipmentRackLoadUseCase`,
  `BulkWeightAdjustDialog`, `AssessmentEngine`); consolidating it with `CommandLimits` is a
  follow-up, not this PR. `HardwareDetection.kt`, also in the list, *was* reconciled in round 1.
- **`spotlessCheck` was not run to green.** It already fails on `main` with violations in 206+
  untouched files and is not part of CI (`ci-tests.yml` runs the two test tasks, `assembleDebug`,
  `:androidApp:lintDebug`, the iOS arm64 compile and the migration verifier).

## Uncompiled iOS changes
- `iosApp/` (SwiftUI) was grepped for `startWorkout` / `sendInitSequence`: **no callers**, so deleting
  the two interface members cannot break the iOS app build.
- `shared/src/iosSimulatorArm64Main/.../PhantomBleRepository.kt` — two `override` keywords removed,
  and (round 1, R-9/R-14/R-43) `ConnectionState.Connected` now carries
  `hardwareModel = HardwareDetection.detectModel(device.name)` plus the import. The phantom
  advertises `Vee_PhantomSimulator`, so it resolves V-Form/100 deliberately instead of Unknown/100
  accidentally.
- `shared/src/iosSimulatorArm64Test/.../PhantomBleRepositoryTest.kt` — three
  `createProgramParams(params)` call sites now pass the newly-required ceiling, plus the import.
- All four are uncompiled on this host **and not compiled by CI** (`ci-tests.yml` builds
  `:shared:compileKotlinIosArm64`, not the simulator target). Everything else is
  `commonMain`/`commonTest`, compiled here for Android.

## Review round 1 (commit `2fe5b5d5`). 0 critical, 5 major, 21 minor, 17 nit — all 43 addressed; per-item Status/Response are in `reviews/pr-13.md`

The five majors were two root causes.

### (A) `WeightChangePerRepControl` reported a programmatic coercion as a user edit
(R-2, R-12, R-13, R-25, R-26, R-29, R-32.2, R-37)

Once the pre-clamps were gone, `_workoutParameters.progressionRegressionKg` held the stored value
(e.g. 5) while the rest card was on screen. The control clamped its own display to ±3 kg and then,
in a `SideEffect`, called `onValueChangeKg(3f)` — **the user-edit handler** — with no user
interaction. Two consequences:
- `updateWorkoutParameters` set `_userAdjustedWeightDuringRest = true`, and the next advance then
  took `currentParams` instead of the routine: set 2 was commanded **set 1's weight and reps**, and
  across an exercise change the previous exercise's weight/reps/echo level/eccentric load. That is
  a wrong-but-legal load, which is exactly what a ceiling clamp cannot catch.
- It rewrote the parameters before the command resolved, so `progressionCapped` was already false
  and the notice never fired in the real app. The acceptance test passed only because the engine
  harness has no Compose runtime.

Fixed at the root, as the reviewers suggested, and without restoring the pre-clamp: **the control is
display-only for out-of-range input.** It renders the clamp and never writes it back;
`CommandLimits.resolve` remains the only thing that changes a commanded value. Its range is now
derived from `CommandLimits.MAX_PROGRESSION_KG` instead of two private constants, which also removes
the lb/kg asymmetry (6.6 lb = 2.99 kg, rather than 6 lb = 2.72 kg) and the pre-existing LB misfire.
The change is one function body plus imports, kept tight because PR 5 edits this file later.

### (B) The capped notice never reached a Just Lift user
(R-1, R-32.1, R-34, R-36)

The notice went out on `_userFeedbackEvents`, a `replay = 0` SharedFlow whose only collector is
`ActiveWorkoutScreen`. Just Lift skips the countdown and navigates there only once the state turns
`Active`, i.e. **after** the send, so `tryEmit` went to zero subscribers and was dropped.

Fixed by making the notice drainable **state**: `WorkoutCoordinator.commandLimitNotice` /
`consumeCommandLimitNotice()`, published by the engine after a successful send and drained by
`ActiveWorkoutScreen` on arrival. The screen shows it on `snackbarScope`, not inside the
`LaunchedEffect` that drains it: draining flips the state to null, which changes the effect's key
and cancels its coroutine, and a cancelled `showSnackbar` dismisses itself — which would have
reproduced the original symptom through a new mechanism. The wiring test asserts the ordering, and
the mutant that moves the call back inside the effect fails exactly that test. That covers every `skipCountdown = true` start, not just Just
Lift, because it no longer depends on a live subscriber at emit time. `_userFeedbackEvents` was
deliberately not given `replay = 1`: it carries unrelated messages that must not re-appear on every
screen entry. The Just Lift weight slider is bounded by the connected model as well, so the common
case no longer needs a notice at all.

### Everything else
- **No fail-open defaults** (R-3, R-16, R-33): `maxWeightPerCableKg` is required on
  `createProgramParams`, `createWorkoutCommand` and `validateLegacyWorkoutCommand`.
- **Echo frames announce nothing** (R-5, R-17): they carry no weight and no progression.
- **Tolerance on both flags** (R-6) and **"(trainer model not recognised)"** wording when the model
  is Unknown (R-20).
- **`HardwareDetection` reconciled** (R-8, R-15, R-31, R-38): the header now says `detectModel` is
  safety-relevant, why the fail-closed direction is acceptable, and records the VIT-200 caution.
  `getCapabilities`/`HardwareCapabilities` (a flat 200 kg, no callers) are deleted.
- **`forceMax` documented** (R-10, R-28, R-30) as firmware force headroom that the per-cable ceiling
  deliberately does not bound, at the assignment and at the test that pins it.
- **`SliderWithButtons`** labels `value.coerceIn(valueRange)` (R-18); **SetReady** falls back to the
  planning ceiling while disconnected (R-42); the dead import is gone (R-7) and the legacy 25-byte
  builder is marked as retained protocol documentation (R-19, R-41).
- **The fake is stricter**: it resolves its model from the device name like production (R-11, R-22),
  denies unrecognised frames by default (R-23), and bounds the Echo frame's eccentric overload
  (R-24).

### Tests added in round 1
- `StoredRoutineLimitsTest.an out of range stored progression does not make set 2 inherit set 1's
  weight and reps` — per-set weights [40, 60] and reps [8, 12] with progression 5; set 2 is
  commanded 60/12 and `_userAdjustedWeightDuringRest` stays false.
- `StoredRoutineLimitsTest.live weight adjustment is bounded by the connected model` (R-21) —
  Trainer+ 115→110, V-Form 105→100, Unknown 105→100, disconnected 105→100.
- New `CommandLimitsWiringTest` — every live and planning slider, the editor progression range, and
  the notice delivery path.
- `RestTimerProgressionWiringTest.weightChangeControlDisplaysTheClampWithoutWritingItBack`
  (replaces `…SyncsClampedDisplayValueBackToParent`).

### Round 1 mutation checks

| Mutant | Failing test |
|---|---|
| The `SideEffect` write-back restored in `WeightChangePerRepControl` | exactly `weightChangeControlDisplaysTheClampWithoutWritingItBack` |
| The notice published on `_userFeedbackEvents` again | `CommandLimitsWiringTest.theCappedNoticeSurvivesTheScreenTransitionThatFollowsTheSend` + 3 `StoredRoutineLimitsTest` cases |
| The Just Lift slider back to the hardcoded 110/242 | exactly `CommandLimitsWiringTest.liveSlidersUseTheConnectedModelCeiling` |

### Honest limitation
The shared suite has no Compose runtime, so the guard for root cause (A) is split: the engine test
pins the *consequence* (set 2 is commanded its own weight and reps, the latch stays false), and a
source-text assertion pins the *cause* (the control has no `SideEffect`). Neither alone would have
caught the original bug; together they do. A Compose UI test would be the real fix and is not in
this PR's scope.

- **The saved session records what was asked, not what was commanded.** `_workoutParameters` still
  holds 105 / 5 after the clamp, and `ActiveSessionEngine.kt:9080` writes
  `WorkoutSession.weightPerCableKg = params.weightPerCableKg` /
  `progressionKg = params.progressionRegressionKg`. The two halves differ, and the round-0 wording
  ("this is not a regression") was only true of one of them:
  - **Weight** — not a regression. Firmware capped at 100 before this PR and the app recorded 105
    then too. What is new is only that the app now *knows* it capped and still records 105.
  - **Progression** — a real change in this PR (R-37). On the paths whose `clampUpcomingProgressionKg`
    pre-clamp was deleted, the session used to record the clamped 3 and now records the stored 5,
    while the frame still carries 3. Round 1 widened this rather than narrowing it: with the
    `WeightChangePerRepControl` write-back gone, `_workoutParameters` keeps the stored value on the
    rendered paths too, not only on the Echo/bodyweight ones.

  Deliberately not fixed here: writing `commandParams` back into `_workoutParameters` would corrupt
  the retry snapshot source, and the plan says stored values must not be rewritten. PR 15 (the atomic
  completion transaction) is the natural place to record the commanded figures — and for progression
  it should be treated as a **fix**, not an open preference. `carryover.md` carries the same wording.
- **Remaining hardcoded 110s upstream of the choke point**, left deliberately because the resolution
  clamp bounds them anyway: `ActiveSessionEngine.kt` `buildWarmupOverrideParams` (~7626) and ~8325,
  and `RoutineFlowManager.kt:1376` (now `Constants.MAX_WEIGHT_PER_CABLE_KG`). These are redundant
  rather than unsafe: each value flows into `bleParams` before resolution. `JustLiftScreen.kt:505`
  was on this list in round 0 and is **no longer** — round 1 bounded that slider by the connected
  model (R-34).
- **Engine user-facing strings are not localized** (pre-existing for every `_userFeedbackEvents` /
  `_bleErrorEvents` message, not just the new one), and the capped notice is always in kg.
- **`Constants.MAX_WEIGHT_KG` (100f) and `MAX_WEIGHT_PER_CABLE_KG` (110f) now overlap with
  `CommandLimits`.** `MAX_WEIGHT_PER_CABLE_KG` is still the right "absolute hardware maximum"
  constant, but the two files should probably be consolidated once no caller needs the model-agnostic
  value.
- **`MonitorDataProcessor` still rejects samples above the model-agnostic 110 kg/cable** (F-060's
  neighbour). Reads, not commands, so out of scope here.
- **Echo frames carry no weight or progression**, so the fake's Echo decode only records the rep
  bytes. Echo's per-cable load bound, if any, is unverified.
- **The `lastConnectedModel` preference is per install, not per trainer.** A user with both models
  gets the last one they connected to for planning.
- **Surface the cap on SetReady before the user taps Start.** Today the weight slider is bounded
  (so a V-Form owner cannot dial past 100), but a stored over-ceiling value and an out-of-range
  stored progression are only explained by the post-send notice. A small "capped for this trainer"
  line on the SetReady card would explain it before the set begins (R-18's second suggestion).
- **A user override for a mis-detected model.** `HardwareDetection` is a name heuristic that the
  per-cable ceiling now depends on. A Trainer+ whose advertised name does not start with `VIT` is
  capped at 100 kg/cable and gets a notice every set, with no way to correct it. A setting, or a
  read of the VERSION characteristic, would fix it; worth a line in `decisions.md` (R-15, R-31).
- **`createWorkoutCommand` sits outside the frame invariant** (R-41): it has no production caller and
  the fake does not decode the 25-byte shape. PR 17 already touches `BlePacketFactory` and could
  either delete it or teach the fake. R-23's deny-by-default `else` arm means a revived caller would
  at least fail the suite rather than pass unnoticed.

## Review
Reviewed by specialist agents at effort 5 (strict: zero open items at every severity). Rounds: 3. Final verdict: pass.


9thLevelSoftware:
### Merge activity

* **Sep 20, 12:09 PM UTC**: @9thLevelSoftware merged this pull request with [Graphite](https://app.graphite.com/github/pr/9thLevelSoftware/Project-Phoenix-MP/821).

